PR #3, opened 2017-08-28
Intial pass at Dockerfile, WIP

efischer19:
Just want to get a branch up to share as needed, nomerge pls

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,15 @@
+FROM node:6.11
+
+USER root
+RUN apt-get update
+
+WORKDIR /edx/prebuilt/studio-frontend
+COPY package.json ./
+RUN npm install --silent
+
+WORKDIR /edx/src/studio-frontend
+ENTRYPOINT cp -r /edx/prebuilt/studio-frontend/node_modules /edx/src/studio-frontend && \
+    npm install && \
+    npm run start
+
+EXPOSE 18123
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,12 @@
+version: "2.1"
+services:
+  studio-frontend:
+    build:
+      context: .
+      dockerfile: Dockerfile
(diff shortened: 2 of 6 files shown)


efischer19:
nvm, there's an official docker image we ought to use: https://github.com/edx/studio-frontend/commit/9f49d6a79018e127e57b5a7b40819ae1d18517e8

docker-compose.yml
```diff
@@ -0,0 +1,11 @@
...
+services:
+  studio-frontend:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    command: bash -c "cd /edx/src/studio-frontend && npm run start"
+    #depends_on:
+    #  - studio
+    volumes:
+      - .:/edx/src/studio-frontend
```
efischer19: To start: `git clean -xfd; docker-compose -f docker-compose.yml up -d studio-frontend`

The `git clean` helps in case the installed `node_modules` gets back into your host machine after a build, that was causing me so much trouble.

Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+RUN apt-get update
+
+WORKDIR /edx/prebuilt/studio-frontend
+COPY package.json ./
+RUN npm install --silent
+
+WORKDIR /edx/src/studio-frontend
+ENTRYPOINT cp -r /edx/prebuilt/studio-frontend/node_modules /edx/src/studio-frontend && \
+    npm install && \
+    npm run start
```
efischer19: This gives us the benefit of a pre-installed `node_modules` folder on each startup, so the following `npm install` should only install new additions to `package.json` that are not yet installed into the image.

efischer19:
This "works" if run independently of studio (obviously, it doesn't proxy anything to studio)

efischer19:
Closing in favor of https://github.com/edx/studio-frontend/pull/5

PR #5, opened 2017-08-29
Dockerization of development environment!

MichaelRoytman:


config/webpack.dev.config.js
```diff
@@ -33,9 +33,12 @@ module.exports = Merge.smart(commonConfig, {
     new webpack.HotModuleReplacementPlugin(),
   ],
   devServer: {
+    host: '0.0.0.0',
+    port: 18011,
+    hot: true,
     proxy: {
       '/api': {
-        target: 'http://localhost:18010',
+        target: 'http://studio:18010',
```
efischer19: I believe we can rewrite this as `http://docker.for.mac.localhost:18010`, per https://docs.docker.com/docker-for-mac/networking/#use-cases-and-workarounds.

We'll need a different approach for prod, but we can cross that bridge when we get there

PR #6, opened 2017-08-29
Use webpack-merge to make a prod webpack config

thallada:
@edx/educator-dahlia

### DIFF
--- a/config/webpack.common.config.js
+++ b/config/webpack.common.config.js
@@ -1,0 +1,23 @@
+'use strict';
+
+const path = require('path');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+
+module.exports = {
+  entry: [
+    path.resolve(__dirname, '../src/index.jsx'),
+  ],
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, '../dist'),
+  },
+  resolve: {
+    extensions: ['.js', '.jsx'],
+  },
+  plugins: [
+    new HtmlWebpackPlugin({
+      inject: true,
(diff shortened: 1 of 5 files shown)


config/webpack.common.config.js
```diff
@@ -0,0 +1,23 @@
...
+  ],
+  output: {
+    filename: 'bundle.js',
+    path: path.resolve(__dirname, '../dist'),
+  },
+  resolve: {
+    extensions: ['.js', '.jsx'],
+  },
+  plugins: [
+    new HtmlWebpackPlugin({
```
arizzitano: We may not want to use this plugin in production if we are just outputting js files for studio's html to consume.

config/webpack.prod.config.js
```diff
@@ -0,0 +1,22 @@
...
+const Merge = require('webpack-merge');
+const commonConfig = require('./webpack.common.config.js');
+const path = require('path');
+const webpack = require('webpack');
+
+module.exports = Merge.smart(commonConfig, {
+  devtool: 'source-map',
+  module: {
+    rules: [
+      {
```
arizzitano: I think we need this rule in the common config, no?

arizzitano: oops it's also in the dev config nvm

config/webpack.common.config.js
thallada: Ah, good point. Also, what do you think about naming the output bundle `studio-frontend.js` in prod?

arizzitano: sounds good to me!

PR #7, opened 2017-08-29
Add assets styles.css and webpack css loaders

thallada:
@edx/educator-dahlia


src/AssetsPage/index.jsx
```diff
@@ -17,11 +20,31 @@ class AssetsPage extends React.Component {
...
-  render() {
+  constructor(props) {
+    super(props);
+    this.state = {
+      checked: 'off',
+    };
     AssetsPage.request();
+  }
+
+  handleCheckBoxChange = (event) => {
```
thallada: I like this style versus having to do `this.handleCheckboxChange = this.handleCheckboxChange.bind(this)` in the constructor. It requires the "transform-class-properties" babel plugin which I added. If you all don't agree let me know.

package.json
```diff
@@ -4,25 +4,37 @@
...
-    "start": "NODE_ENV=development BABEL_ENV=development webpack-dev-server --config=config/webpack.dev.config.js --progress",
+    "start": "NODE_ENV=development BABEL_ENV=development webpack-dev-server --config=config/webpack.dev.config.js --hot --inline --progress",
     "build": "NODE_ENV=production BABEL_ENV=production webpack --config=config/webpack.prod.config.js",
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "license": "AGPL-3.0",
   "dependencies": {
+    "bootstrap": "^4.0.0-beta",
+    "edx-bootstrap": "^0.1.7",
+    "paragon": "git://github.com/edx/paragon.git",
```
thallada: Waiting on paragon getting published on the edx npm org.

src/AssetsPage/index.jsx
```diff
@@ -17,11 +20,31 @@ class AssetsPage extends React.Component {
...
+    console.log(`checked: ${event}`);
+    this.setState({
+      checked: event,
+    });
```
thallada: The checkbox is kind of broken right now so this doesn't really do anything. See https://github.com/edx/paragon/pull/26

config/webpack.dev.config.js
```diff
@@ -8,21 +8,43 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 module.exports = Merge.smart(commonConfig, {
   devtool: 'cheap-module-eval-source-map',
-  entry: [
-    `${require.resolve('webpack-dev-server/client')}?/`,
-    require.resolve('webpack/hot/dev-server'),
-    require.resolve('react-error-overlay'),
-  ],
```
thallada: I'm able to remove this because the `--hot --inline` options to the webpack-dev-server does this for us automatically.

thallada:
If you have previously checked this branch out, you will have to `rm -rf node_modules/paragon` and then `npm install` again to pick up the paragon CheckBox fixes that this now relies on.

.eslintrc
```diff
@@ -0,0 +1,3 @@
+{
+  "parser": "babel-eslint"
```
arizzitano: I already added eslint config [as `eslintConfig` within package.json](https://github.com/edx/studio-frontend/blob/master/package.json#L49-L61), which is what create react app does. I was thinking that would be an easier central location than an .eslintrc, but now that our config is growing in size, I'm thinking it might be better to just dump it all into this eslintrc. What do you think @thallada ?

thallada: Oh I didn't even see that. I can move that config here. Also, I think I'm going to rename this to `.eslintrc.json` selfishly just because it makes my editor know to syntax highlight it as json.

PR #8, opened 2017-08-30
Run docker builds in travis and add some make commands

MichaelRoytman:


config/webpack.dev.config.js
```diff
@@ -33,9 +33,11 @@ module.exports = Merge.smart(commonConfig, {
     new webpack.HotModuleReplacementPlugin(),
   ],
   devServer: {
+    host: '0.0.0.0',
+    port: 18011,
     proxy: {
       '/api': {
-        target: 'http://localhost:18010',
+        target: 'http://docker.for.mac.localhost:18010',
```
thallada: I can't fully test it because my devstack is broken right now (going to destroy and start again), but I'm pretty sure this won't work on my ubuntu machine :penguin:

MichaelRoytman:
Please let me know which make commands, if any, would be useful to keep. I just put in some I thought might be useful, but I realize some may be redundant or unnecessary.

Makefile
```diff
@@ -1,2 +1,38 @@
...
+# Help message borrowed from https://github.com/edx/devstack, which borrowed it from https://github.com/pydanny/cookiecutter-djangopackage.
+help: ## display a help message
+	@echo $(MAKEFILE_LIST)
+	@echo "Please use \`make <target>' where <target> is one of"
+	@perl -nle'print $& if m{^[\.%a-zA-Z_-]+:.*?## .*$$}' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m  %-25s\033[0m %s\n", $$1, $$2}'
+
+shell: ## run a shell on the studio-frontend container
 	docker exec -it dahlia.studio-frontend /bin/bash
+
+dev.up: ## bring up studio-frontend container
```
efischer19: Personal opinion: I'd rather save the extra keystrokes and just have `up` instead of `dev.up` for all of these commands

Makefile
```diff
@@ -1,2 +1,38 @@
...
+	docker-compose up -d studio-frontend
+
+dev.down: ## stop and remove studio-frontend container
+	docker-compose down studio-frontend
+
+dev.stop: ## stop studio-frontend container but do not remove
+	docker-compose stop studio-frontend
+
+dev.start: ## start stopped studio-frontend container
+	docker-compose start studio-frontend
```
efischer19: How exactly do `up`, `build`, and `start` compare? I feel like we don't need all of them, but I'm not 100% on the difference. I usually only use up (which is similar to build+start?)

I'd also like to request a single command for "stop, remove, rebuild (without cache), and restart"

MichaelRoytman: They're not terribly different, and we could definitely roll them up into one if that's how we're most likely to use them. The way I understand it is:

- dev.up = using an existing image tagged edxops/studio-frontend (or pulling down from docker hub), start the container

- dev.start = restart an existing container that was brought up and then stopped, since dev.down actually removes the container; I guess this would save us the step of having to rerun the entrypoint command the way we would if we did dev.down and dev.up, but I'm not sure how big of a gain that is

- dev.build = thinking more about it, I don't think it's super useful, since we have the bind mounting. I thought maybe it would be useful to either re-build an image locally that differs from the one on DockerHub (with same tag) or to build an image with a different tag for testing purposes. Could there be local changes that don't affect the image until the image is rebuilt? For example, a change to the Dockerfile wouldn't affect the running container, and you'd want to test it out by rebuilding. The re-tagging would allow you to keep the existing image without overwriting it.

Does that help us determine what would be useful? I'm fine removing any that seem superfluous. 

And that's a good idea. I'll add some kind of start from scratch command.

Makefile
MichaelRoytman: That's a good idea. I found it cumbersome too. I'll fix that.

.travis.yml
```diff
@@ -3,5 +3,12 @@ language: node_js
 node_js:
   - 6
 
+sudo: required
+services:
+  - docker
+
+before_script:
+  - docker build -t studio-frontend --no-cache .
```
MichaelRoytman: This builds the image before the script runs, so if we wanted to run tests on the image, we could do it in the script step, when the time comes.

Makefile
```diff
@@ -1,2 +1,32 @@
...
+	docker-compose logs -f studio-frontend
+
+down: ## stop and remove studio-frontend container
+	docker-compose down
+
+npm-install-%: ## install specified % npm package on the studio-frontend container
+	docker exec dahlia.studio-frontend npm install $* --save-dev
+	git add package.json
+
+from-scratch: ## start development environment from scratch
```
efischer19: Super big fan of this one, BTW

PR #9, opened 2017-08-30
Handle no studio

efischer19:
FYI @edx/educator-dahlia 

Shoutout to @arizzitano helping me get this sorted out.

### DIFF
--- a/src/AssetsPage/index.jsx
+++ b/src/AssetsPage/index.jsx
@@ -4,28 +4,11 @@ import CheckBox from 'paragon/src/CheckBox';
 import styles from './styles.scss';
 
 class AssetsPage extends React.Component {
-  static request() {
-    fetch('/api/assets/course-v1:edX+DemoX+Demo_Course/?page=0&page_size=50&sort=sort&asset_type=', {
-      credentials: 'same-origin',
-      headers: {
-        Accept: 'application/json',
-      },
-    })
-      .then(data => data.json())
-      .catch(() => {
-        // console.log(error);
-      })
-      .then(() => {
-        // console.log(json);
-      });
-  }
-
   constructor(props) {
     super(props);
     this.state = {
       checked: false,
     };
-    AssetsPage.request();
   }
 
   handleCheckBoxChange = (checked) => {
--- a/src/BackendStatusBanner/index.jsx
+++ b/src/BackendStatusBanner/index.jsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import statusMap from './statusMap.json';
+
+class BackendStatusBanner extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      apiConnectionStatus: 200,
+    };
+  }
+
+  componentDidMount() {
+    fetch('/api/assets/course-v1:edX+DemoX+Demo_Course/?page=0&page_size=50&sort=sort&asset_type=', {
+      credentials: 'same-origin',
+      headers: {
+        Accept: 'application/json',
+      },
+    })
+      .then((response) => {
+        if (response.status !== this.state.apiConnectionStatus) {
+          this.setState({
+            apiConnectionStatus: response.status,
+          });
+        }
+      });
+  }
+
+  render() {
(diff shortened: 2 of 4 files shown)


src/AssetsPage/index.jsx
```diff
@@ -1,26 +1,37 @@
...
   render() {
-    AssetsPage.request();
+    this.request();
+    const backendStatusDiv =
+        (this.apiConnectionStatus === 200) ?
+          null :
+          (<div className="api-error">{statusMap[this.state.apiConnectionStatus]}</div>);
     return (
       <div>
+        {backendStatusDiv}
```
thallada: I think this should be made into a separate component because it that seems like the general best practice thing to do in React and I can definitely see reusing this in our other pages. I'm imagining:

```jsx
import BackendStatus from './BackendStatus';
...
render() {
  return (
    <div>
      <BackendStatus
        statusCode={this.state.apiConnectionStatus}
      />
      <h2>I am the AssetsPage</h2>
    </div>
  );
}
```

Then, BackendStatus receives the status code as a prop and could handle the logic of mapping the code to the message and how to display it.

src/AssetsPage/index.jsx
```diff
@@ -1,26 +1,37 @@
...
     })
-      .then(data => data.json())
-      .catch(() => {
-        // console.log(error);
-      })
-      .then(() => {
-        // console.log(json);
+      .then((response) => {
+        if (response.status !== this.state.apiConnectionStatus) {
+          this.setState(() => ({ apiConnectionStatus: response.status }));
```
thallada: FYI you can just do `this.setState({apiConnectionStatus: response.status});` here. It will shallow merge that object into the current state.

src/AssetsPage/index.jsx
```diff
@@ -1,26 +1,37 @@
...
+        }
       });
   }
 
   render() {
-    AssetsPage.request();
+    this.request();
```
thallada: This function might call `setState` which will trigger a re-render, causing this component to be rendered two times in a row initially if the API request fails. Maybe this call should go in the constructor. I don't think it needs to be called every render, just one time initially for now.

src/AssetsPage/statusMap.json
```diff
@@ -0,0 +1,5 @@
+{
+  "504": "Studio is not running",
+  "404": "You are not logged in to Studio",
+  "403": "You are logged in to Studio, but your account doesn't have the correct permissions"
```
thallada: Wouldn't hurt to define 500 here too so we know we broke API code when we are developing on the assets page later.

src/AssetsPage/index.jsx
arizzitano: Tyler makes a good point about componentizing when we can. I'd take a slightly different approach to this component, though. We know the following things:
- we want to show this status banner on all pages within our development environment
- we do *not* want to show or even include this status banner in our production builds
- this status banner should (probably) always show up in the same place
- if we are tackling this project as a single page app, we will need to set up routing at the root layer of the app
- the status banner doesn't really have anything to do with the assets page specifically

Based on all this, I'd do the following:
- Create a new component `BackendStatusBanner` or something along those lines. 
- Move all the functionality you've implemented here (the fetch call, the status mapping, etc) into your new component.
- Include the component not within `AssetsPage` (which will someday soon contain our lovely UI for the new assets page) but at the root level, within `App`, which is defined one level up in `src/index.jsx`.

It would look something like:
```
class BackendStatusBanner extends React.Component {
    // put basically everything from AssetsPage in here except for the "I am an AssetsPage" bit
}

const App = () => (
  <div>
    <BackendStatusBanner />
    <h1>I am a app</h1>
    <AssetsPage />
  </div>
);
```

This way, if we export `AssetsPage` for production as an individual module to be used within studio, it won't have any reference to the backend status, and once we set up routing at the root level (replacing the `<AssetsPage/>` in `App` with a more sophisticated route handler), the backend status banner will show up on every page within our development environment.

Just my take though, and super happy to discuss further!

PR #10, opened 2017-08-30
Mac + Linux = Best Buddies

efischer19:


thallada:
It works for me! :confetti_ball: Thanks for fixing it!

efischer19:
Awesome! I'll wait until https://github.com/edx/studio-frontend/pull/8 merges since this actually relies on it, then get this in ASAP.

docker-compose.linux.yml
```diff
@@ -0,0 +1,7 @@
+studio-frontend-linux:
+    extends:
+        file: docker-compose.yml
+        service: studio-frontend
+    environment:
+        - RUNNING_ON_LINUX=1
+    network_mode: host
```
efischer19: Add a comment in this file before merging.

"This works because docker-compose will merge the files together. Mac host machines cannot publish ports in network_mode='host', Linux users require network_mode='host' because there's no equivalent to docker.for.mac.localhost for that OS."

thallada:
Actually, I lied, sorry :cry:. I realized I didn't actually test this. I am running into some issues with this that I'll have to ask you about tomorrow morning.

PR #11, opened 2017-08-31
(WIP) Extract Display Strings

efischer19:


efischer19:
Spoke with @thallada offline about this - we're worried that there's going to be a lot of overhead in completely overhauling the i18n system, when edx-platform already has a fairly mature pipeline in place. I'm going to pivot to investigating how hard it would be to link up with that process.

efischer19:
Relevant links:

- [edx-platform defining ignore directories](https://github.com/edx/edx-platform/blob/41f7502ae37873f3d59af4a0c95c2a1c83705d8d/conf/locale/config.yaml#L98)
- [where our i18n tools checks those config values](https://github.com/edx/i18n-tools/blob/ee649fe87014c3948d3eec25e9580f7e6c9330c8/i18n/extract.py#L120)

efischer19:
This was a premature PR, none of this work is needed any longer.

PR #12, opened 2017-09-01
[WIP] display error output in browser

arizzitano:
Playing with [react-error-overlay](https://github.com/facebookincubator/create-react-app/blob/70768b321e357c56bfb4cdc52be0461ddd7cbadc/packages/react-error-overlay/README.md) to make compile or execution errors display in the browser. It leverages react-devtools' alternative hot reloading client. Currently an error looks like this:

image

CRA has an easier-to-read "dark" theme though so I'm trying to force that one to show up instead.

### DIFF
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -5,7 +5,6 @@ const commonConfig = require('./webpack.common.config.js');
 const path = require('path');
 const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
-const errorOverlayMiddleware = require('react-error-overlay/middleware');
 
 let targetUrl = 'localhost';
 if (!process.env.RUNNING_ON_LINUX) {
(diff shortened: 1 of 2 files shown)


config/webpack.dev.config.js
```diff
@@ -69,5 +70,13 @@ module.exports = Merge.smart(commonConfig, {
...
       },
     },
+    setup(app) {
+      app.use(errorOverlayMiddleware());
+    },
   },
+  entry: [
+    require.resolve('react-dev-utils/webpackHotDevClient'),
+    require.resolve('react-error-overlay'),
+    path.resolve(__dirname, '../src/index.jsx'),
```
thallada: I think this line is unnecessary because this will be merged with the entry array in [`webpack.common.config.js`](https://github.com/edx/studio-frontend/blob/69c5a19737a9b65e90837ebadb80e42291a937c7/config/webpack.common.config.js#L7).

thallada: Oh wait, that's not an array. Maybe it won't merge, not sure actually...

arizzitano: Nope, Merge can't handle an object + an array. It actually sort of works merging with the following in webpack.dev.config:
```
  entry: {
    hotDev: require.resolve('react-dev-utils/webpackHotDevClient'),
  },
```

But that makes the hot reloader occasionally fill up the call stack for some reason :/

PR #15, opened 2017-09-01
data layer

### LINKED ISSUE #14: data layer

arizzitano:
closes #14 

- install redux, redux-thunk, react-redux, and prop-types
- set up a redux store to be used throughout the entire studio-frontend application
    - currently it only has one reducer `connectionStatus`, but it is ready for more!
- modify @efischer19's `BackendStatusBanner`
    - wire it up to redux
    - move the XHR into a redux action
    - add some `propTypes` and `defaultProps`

I was thinking about moving the `BackendStatusBanner` display text into redux as well, what do yall think? I can't really decide whether that text is UI (and should live in the component) or state (and should live in the data layer).

### DIFF
--- a/src/BackendStatusBanner/index.jsx
+++ b/src/BackendStatusBanner/index.jsx
@@ -3,7 +3,7 @@ import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 
 import statusMap from './statusMap.json';
-import { pingStudio } from '../data/store';
+import { pingStudio } from '../data/actions/pingStudio';
 
 class BackendStatusBanner extends React.Component {
   constructor(props) {
@@ -18,24 +18,28 @@ class BackendStatusBanner extends React.Component {
   }
 
   render() {
-    return (this.apiConnectionStatus === 200) ?
+    return (this.props.connectionStatus === 200) ?
       null :
       (
         <div className="api-error">
-          {statusMap[this.props.apiConnectionStatus]}
+          {statusMap[this.props.connectionStatus]}
         </div>
       );
   }
 }
 
 BackendStatusBanner.propTypes = {
-  apiConnectionStatus: PropTypes.string.isRequired,
+  connectionStatus: PropTypes.number,
   pingStudio: PropTypes.func.isRequired,
+};
+
+BackendStatusBanner.defaultProps = {
+  connectionStatus: null,
 };
 
 const WrappedBackendStatusBanner = connect(
   state => ({
-    apiConnectionStatus: state.apiConnectionStatus,
+    connectionStatus: state.connectionStatus,
   }), dispatch => ({
     pingStudio: () => dispatch(pingStudio()),
   }),
--- a/src/BackendStatusBanner/index.jsx
+++ b/src/BackendStatusBanner/index.jsx
@@ -1,5 +1,9 @@
 import React from 'react';
+import { connect } from 'react-redux';
+import PropTypes from 'prop-types';
+
 import statusMap from './statusMap.json';
+import { pingStudio } from '../data/store';
 
 class BackendStatusBanner extends React.Component {
   constructor(props) {
@@ -7,27 +11,10 @@ class BackendStatusBanner extends React.Component {
     this.state = {
       apiConnectionStatus: 200,
     };
-
-    this.apiUrl = '/assets/course-v1:edX+DemoX+Demo_Course/?page=0&page_size=50&sort=sort&asset_type=';
-    if (!process.env.NODE_ENV || process.env.NODE_ENV === 'development') {
-      this.apiUrl = `/api${this.apiUrl}`;
-    }
   }
 
   componentDidMount() {
-    fetch(this.apiUrl, {
-      credentials: 'same-origin',
-      headers: {
-        Accept: 'application/json',
-      },
-    })
-      .then((response) => {
-        if (response.status !== this.state.apiConnectionStatus) {
-          this.setState({
-            apiConnectionStatus: response.status,
-          });
-        }
-      });
+    this.props.pingStudio();
   }
 
   render() {
@@ -35,10 +22,23 @@ class BackendStatusBanner extends React.Component {
       null :
       (
         <div className="api-error">
-          {statusMap[this.state.apiConnectionStatus]}
+          {statusMap[this.props.apiConnectionStatus]}
         </div>
       );
   }
 }
 
-export default BackendStatusBanner;
+BackendStatusBanner.propTypes = {
+  apiConnectionStatus: PropTypes.string.isRequired,
+  pingStudio: PropTypes.func.isRequired,
+};
+
+const WrappedBackendStatusBanner = connect(
+  state => ({
+    apiConnectionStatus: state.apiConnectionStatus,
+  }), dispatch => ({
+    pingStudio: () => dispatch(pingStudio()),
+  }),
+)(BackendStatusBanner);
+
+export default WrappedBackendStatusBanner;
--- a/src/data/actions/pingStudio.js
+++ b/src/data/actions/pingStudio.js
@@ -7,10 +7,9 @@ export const pingResponse = response => ({
   status: response.status,
 });
 
-export const pingStudio = () => {
-  return dispatch =>
+export const pingStudio = () =>
+  dispatch =>
     getAssets('course-v1:edX+DemoX+Demo_Course', {
       page: 0,
     })
       .then(response => dispatch(pingResponse(response)));
-};
--- a/src/data/actions/pingStudio.js
+++ b/src/data/actions/pingStudio.js
@@ -1,0 +1,19 @@
+export const PING_RESPONSE = 'PING_RESPONSE';
+
+function pingResponse(response) {
+  return {
+    type: PING_RESPONSE,
+    status: response.status,
+  };
+}
+
+export function pingStudio() {
+  return dispatch =>
+    fetch('/api/assets/course-v1:edX+DemoX+Demo_Course/?page=0&page_size=50&sort=sort&asset_type=', {
+      credentials: 'same-origin',
+      headers: {
+        Accept: 'application/json',
+      },
+    })
+      .then(response => dispatch(pingResponse(response)));
+}
--- a/src/data/actions/pingStudio.js
+++ b/src/data/actions/pingStudio.js
@@ -2,17 +2,15 @@ import { getAssets } from '../../api/client';
 
 export const PING_RESPONSE = 'PING_RESPONSE';
 
-function pingResponse(response) {
-  return {
-    type: PING_RESPONSE,
-    status: response.status,
-  };
-}
+export const pingResponse = response => ({
+  type: PING_RESPONSE,
+  status: response.status,
+});
 
-export function pingStudio() {
+export const pingStudio = () => {
   return dispatch =>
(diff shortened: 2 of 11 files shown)


src/data/actions/pingStudio.js
```diff
@@ -0,0 +1,19 @@
+export const PING_RESPONSE = 'PING_RESPONSE';
+
+function pingResponse(response) {
```
arizzitano: This is a vanilla, non-asynchronous action creator, returning a very simple action (like a message for the queue) containing response status. The `PING_RESPONSE` label is important -- we'll see why later. Docs on actions and action creators available here: http://redux.js.org/docs/basics/Actions.html

src/data/actions/pingStudio.js
```diff
@@ -0,0 +1,19 @@
...
+  return {
+    type: PING_RESPONSE,
+    status: response.status,
+  };
+}
+
+export function pingStudio() {
```
arizzitano: This is an asynchronous action creator, also known as a thunk. Action creator logic is polymorphic, so if it returns an object (like `pingStudio` above) it is treated as synchronous. However, if, like this one, it returns a function, it is treated as asynchronous. http://redux.js.org/docs/advanced/AsyncActions.html

src/data/actions/pingStudio.js
```diff
@@ -0,0 +1,19 @@
...
+  return dispatch =>
```
arizzitano: Look who it is -- our old friend `dispatch`!

src/data/actions/pingStudio.js
```diff
@@ -0,0 +1,19 @@
...
+    fetch('/api/assets/course-v1:edX+DemoX+Demo_Course/?page=0&page_size=50&sort=sort&asset_type=', {
+      credentials: 'same-origin',
+      headers: {
+        Accept: 'application/json',
+      },
+    })
+      .then(response => dispatch(pingResponse(response)));
```
arizzitano: Like in our `mapDispatchToProps` defined on `BackendStatusBanner`, we use `dispatch` to invoke an action creator. That's pretty much all `dispatch` does, and it's important that we only invoke action creators through `dispatch` so their resulting data is handled through Redux, without unexpected side effects. This is the power of one-way data flow -- single source of truth, no funky rando jQuery functions mutating data that's inaccessible to other UI. 

Within the `fetch` call's asynchronous promise chain, we invoke `pingResponse` passing in the HTTP response, so it can then be digested and stored by Redux.

src/data/reducers/connectionStatus.js
```diff
@@ -0,0 +1,10 @@
+import { PING_RESPONSE } from '../actions/pingStudio';
+
+export default function connectionStatus(state = null, action) {
```
arizzitano: This function is a reducer (http://redux.js.org/docs/basics/Reducers.html), the final step in the data flow process. Anything it returns gets stored as application state. Its name `connectionStatus` becomes the root-level key in the Redux store, so anything this function returns is stored under `connectionStatus`.

src/data/store.js
```diff
@@ -0,0 +1,11 @@
+import { applyMiddleware, createStore } from 'redux';
+import thunkMiddleware from 'redux-thunk';
+
+import rootReducer from './reducers';
+
+const store = createStore(
+  rootReducer,
+  applyMiddleware(thunkMiddleware),
```
arizzitano: Thunks aren't supported out of the box -- we have to enable them via middleware.

src/data/store.js
```diff
@@ -0,0 +1,11 @@
+import { applyMiddleware, createStore } from 'redux';
+import thunkMiddleware from 'redux-thunk';
+
+import rootReducer from './reducers';
+
+const store = createStore(
+  rootReducer,
```
arizzitano: We pass our `rootReducer` in here so the store knows where its data is coming from.

src/data/reducers/index.js
```diff
@@ -0,0 +1,9 @@
+import { combineReducers } from 'redux';
+
+import connectionStatus from './connectionStatus';
+
+const rootReducer = combineReducers({
```
arizzitano: `combineReducers` multiplexes together all the reducers in the application (although currently we just have one) so they can be easily passed into the store.

src/data/reducers/connectionStatus.js
```diff
@@ -0,0 +1,10 @@
...
+  switch (action.type) {
+    case PING_RESPONSE:
```
arizzitano: `connectionStatus` (and all other reducers in the application) is invoked every time an action creator passes data through Redux via an action -- essentially, every time a message hops on the bus. This allows it to pull in whatever action data it cares about, digest it, and return it. 

This is all very well and good, but the reducer probably doesn't care about every single action passing through -- it really only cares about the connection status. How can it pull out only what it needs? Well, remember `PING_RESPONSE` from our `pingStudio` action creator? This reducer listens for actions of that `type` so it can pull out their data and digest it. It's Redux convention to use a switch statement to accomplish this. If we wanted to, we could add other action type slugs to listen for, but we're just keeping it simple for now.

src/index.jsx
```diff
@@ -1,15 +1,19 @@
...
 import BackendStatusBanner from './BackendStatusBanner';
+import store from './data/store';
 
 const App = () => (
-  <div>
-    <BackendStatusBanner />
-    <h1>I am a app</h1>
-    <AssetsPage />
-  </div>
+  <Provider store={store}>
```
arizzitano: Finally, the last step. In order to make all the data in the store accessible to the rest of the application, we need to wrap it in a `react-redux` Provider at the root level. This isn't really something we'll have to worry about ever again.

src/BackendStatusBanner/index.jsx
```diff
@@ -1,39 +1,41 @@
...
         <div className="api-error">
-          {statusMap[this.state.apiConnectionStatus]}
+          {statusMap[this.props.connectionStatus]}
         </div>
       );
   }
 }
 
-export default BackendStatusBanner;
+BackendStatusBanner.propTypes = {
```
arizzitano: `propTypes` is a neat module that enables us to do type checking on the props (props are basically arguments) passed into the component. https://github.com/facebook/prop-types

These props can be passed where the component gets invoked (e.g. `<BackendStatusBanner connectionStatus={200} />, or they can be indirectly passed through Redux, as we are doing below.

src/BackendStatusBanner/index.jsx
```diff
@@ -1,39 +1,41 @@
...
+  connectionStatus: PropTypes.number,
+  pingStudio: PropTypes.func.isRequired,
+};
+
+BackendStatusBanner.defaultProps = {
```
arizzitano: For any prop types that aren't `isRequired`, we provide a default value. `connectionStatus` is null in Redux until its XHR returns, so we indicate its default should be `null` here also.

src/data/reducers/connectionStatus.js
MichaelRoytman: While watching the egghead tutorial, I saw Dan Abramov using the ES6 arrow function notation for his reducer functions. Is there an advantage to doing it that way? Would it be better to use that notation, or is there a reason to keep it this way? Just curious!

src/data/reducers/connectionStatus.js
```diff
@@ -0,0 +1,10 @@
...
+      return action.status;
+    default:
+      return state;
```
dsjen: Yes! The tutorial said to do this. :)

src/BackendStatusBanner/index.jsx
```diff
@@ -1,39 +1,41 @@
...
+  connectionStatus: null,
+};
+
+const WrappedBackendStatusBanner = connect(
+  state => ({
+    connectionStatus: state.connectionStatus,
+  }), dispatch => ({
+    pingStudio: () => dispatch(pingStudio()),
```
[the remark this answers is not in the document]
MichaelRoytman: This is more a question for my own understanding than a comment on the code. Is this mapping the pingStudio function that's being used at this [line](https://github.com/edx/studio-frontend/pull/15/files#diff-7fadfae4565e9c1514e2a49e17a2288cR17) to the function dispatch(pingStudio())? Is the function considered a thunk - not sure if I'm using thunk correctly in this context, since I'm leaning a bit on my knowledge of Scheme? Is its execution delayed until it's used from the aforementioned line 17?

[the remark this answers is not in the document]
arizzitano: @MichaelRoytman exactly. The arrow function `() => dispatch(pingStudio())` wraps the invocation of the action creator `pingStudio` (the one we import at the top of the file) in a function. So when this function is called, it invokes `pingStudio` and passes its result to `dispatch`. Like you said, it only gets called where the prop function is invoked on line 17. 

One thing I want to point out is that the prop name `pingStudio` (the key in the object containing this function) does *not* have to match the name of the action creator. We could do
```
foo: () => dispatch(pingStudio()),
```
and then within our component we'd invoke `this.props.foo()`, and it would do the same thing. I just called it `pingStudio` to be consistent.

src/BackendStatusBanner/index.jsx
```diff
@@ -1,44 +1,48 @@
...
+  }),
+)(BackendStatusBanner);
+
+export default WrappedBackendStatusBanner;
```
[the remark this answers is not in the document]
arizzitano: Good point! I wish we could, but you can't `export default const` ([this answer gives a pretty good explanation why](https://stackoverflow.com/a/36261387/1135244)). I'd be inclined to do something like
```
export default WrappedBackendStatusBanner = connect(
```
But because `WrappedBackendStatusBanner` contains a *class* and not a *function*, that syntax doesn't quite work. We'd have to declare an empty `let WrappedBackendStatusBanner;` further up the file, which I'm not personally a huge fan of.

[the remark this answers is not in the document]
arizzitano: Errata: I'm wrong -- `WrappedBackendStatusBanner` does actually contain a function. It's the type of declaration (LexicalDeclaration) that makes things tricky here. That SO answer explains it much better than I can.

src/data/reducers/connectionStatus.js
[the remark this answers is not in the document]
arizzitano: @dsjen `export default` works inline here because `connectionStatus` is a named function declared as a `FunctionDeclaration`, whereas `WrappedBackendStatusBanner` is declared as a LexicalDeclaration. Basically, if it contains an = sign or could be assigned to a variable, it's a LexicalDeclaration and has to be exported default separately. If it can't be assigned to a variable (e.g. `class MyClass {`, `function myFunction`, etc) it's a HoistableDeclaration, ClassDeclaration or AssignmentExpression, and we can export default it inline.

@MichaelRoytman that's a good point. Now, if we change this function to an arrow function instead of a named function, we end up with something like `const connectionStatus = () => {`. That turns it into a LexicalDeclaration, and we'll have to add an `export default connectionStatus;` at the bottom of the file like with the `WrappedBackendStatusBanner`. I'm going to make that change, at least for consistency's sake -- as long as we're cool with `export default`s at the bottom of files. (I am!)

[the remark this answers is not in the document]
dsjen: Sounds like a plan! I appreciate you taking the time to explain what's going on here, @arizzitano.

PR #16, opened 2017-09-05
Set NODE_ENV to production in prod & change API url in prod

thallada:
This is to prepare the production build for importing and embedding into the Studio Files & Uploads page (PR coming soon).

The React Dev Tools yells at me unless I define the `process.env.NODE_ENV` as `production` (thanks @arizzitano for suggesting that extension) ([react docs on why](https://facebook.github.io/react/docs/optimizing-performance.html#use-the-production-build)) which is why I added the `DefinePlugin` to the prod webpack config. I also use that env var to change the assets API url in production at run-time, which we prepended with `/api` in development.

@edx/educator-dahlia

### DIFF
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -51,6 +51,11 @@ module.exports = Merge.smart(commonConfig, {
   },
   plugins: [
     new ExtractTextPlugin('studio-frontend.min.css'),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify('production'),
+      },
+    }),
     new webpack.optimize.UglifyJsPlugin({ sourceMap: true }),
   ],
 });
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -53,7 +53,7 @@ module.exports = Merge.smart(commonConfig, {
     new ExtractTextPlugin('studio-frontend.min.css'),
     new webpack.DefinePlugin({
       'process.env': {
-        NODE_ENV: JSON.stringify('production'),
+        NODE_ENV: JSON.stringify(process.env.NODE_ENV),
       },
     }),
     new webpack.optimize.UglifyJsPlugin({ sourceMap: true }),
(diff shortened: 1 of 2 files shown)


src/BackendStatusBanner/index.jsx
```diff
@@ -7,10 +7,15 @@ class BackendStatusBanner extends React.Component {
     this.state = {
       apiConnectionStatus: 200,
     };
+
+    this.apiUrl = '/assets/course-v1:edX+DemoX+Demo_Course/?page=0&page_size=50&sort=sort&asset_type=';
+    if (!process.env.NODE_ENV || process.env.NODE_ENV === 'development') {
+      this.apiUrl = `/api${this.apiUrl}`;
+    }
```
arizzitano: This is a fine workaround for now, but I'll probably change it in my redux PR. We discussed this a bit last week -- ultimately, if we store all our Studio URLs in a separate manifest file, we can import them directly into the `devServer` config, not have to worry about pattern matching the `/api` slug, and use the list of URLs as a kind of contract with Studio.

config/webpack.prod.config.js
```diff
@@ -51,6 +51,11 @@ module.exports = Merge.smart(commonConfig, {
   },
   plugins: [
     new ExtractTextPlugin('studio-frontend.min.css'),
+    new webpack.DefinePlugin({
+      'process.env': {
+        NODE_ENV: JSON.stringify('production'),
```
arizzitano: If `NODE_ENV=production` is set in the production build task (looks like it is), can you pull in `process.env.NODE_ENV` here instead of hardcoding the value?

src/BackendStatusBanner/index.jsx
thallada: That makes sense and I'd be happy to do that work, but it probably makes sense to do that on top of your redux work. Are you okay if I merge this as it is now and you can copy this block into the reducer when you fix the conflict? Then, I can make a PR off your redux one that creates a separate manifest file with our API urls and import them into the reducer and our webpack dev config.

arizzitano: For sure, go for it! Don't sweat the conflicts -- I was planning on just addressing them anyway.

PR #17, opened 2017-09-06
publish to npm

arizzitano:
This will enable us to publish the package to npm.

cc @edx/educator-dahlia


arizzitano:
Good call @thallada, done!

PR #18, opened 2017-09-06
API client

thallada:
Created `getAssets` which abstracts away creating a fetch request for the `pingStudio` action (and any actions we make in the future).

All API endpoints (top-level) we use should be listed in `src/api/endpoints.js` which the webpack dev config will use to create devServer proxies.

I've never done this before, so this is just my best guess at the best pattern. I looked at stuff like [redux-rest](https://github.com/Kvoti/redux-rest) but I either didn't get it or thought it was way too complicated for our purposes.

### DIFF
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -14,7 +14,7 @@ if (!process.env.RUNNING_ON_LINUX) {
 }
 targetUrl = `http://${targetUrl}:18010`;
 
-const devConfig = Merge.smart(commonConfig, {
+module.exports = Merge.smart(commonConfig, {
   devtool: 'cheap-module-eval-source-map',
   entry: [
     // enable react's custom hot dev client so we get errors reported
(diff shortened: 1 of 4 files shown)


config/webpack.dev.config.js
```diff
@@ -69,11 +71,14 @@ module.exports = Merge.smart(commonConfig, {
...
-      },
-    },
+    proxy: {},
   },
 });
+
+Object.keys(apiEndpoints).forEach((endpoint) => {
+  devConfig.devServer.proxy[apiEndpoints[endpoint]] = {
+    target: targetUrl,
+  };
```
arizzitano: I actually don't think you need the `target` unless you're doing a `pathRewrite`, so an object like
```
{
    <apiEndpoint1>: targetUrl,
    <apiEndpoint2>: targetUrl,
    // ...
}
```
should be fine. You could also do this inline as part of the config itself with a `map` statement, e.g.
```
proxy: Object.keys(apiEndpoints).map(endpoint => apiEndpoints[endpoint])
```

ES2015 FTW!

PR #20, opened 2017-09-07
display assets data in a table

arizzitano:
* set up a new reducer `assetsList`, distinct from the one pinging for status, to store the list of assets fetched from the api
* wire up `AssetsPage` to redux to fire the appropriate action and listen to the appropriate `assetsList` store slice
* add a new `AssetsTable` component and wire it into the `AssetsPage`
* use the paragon `Table` component within `AssetsTable` (https://github.com/edx/paragon/pull/32/)

@edx/educator-dahlia

### DIFF
--- a/src/AssetsPage/AssetsTable/index.jsx
+++ b/src/AssetsPage/AssetsTable/index.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import Table from 'paragon/src/Table';
 
-const AssetsTable = ({assetsList}) => (
+const AssetsTable = ({ assetsList }) => (
   (!assetsList.length) ? (
     <span>Loading....</span>
   ) : (
--- a/src/AssetsPage/AssetsTable/index.jsx
+++ b/src/AssetsPage/AssetsTable/index.jsx
@@ -1,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Table from 'paragon/src/Table';
+
+const AssetsTable = ({assetsList}) => (
+  (!assetsList.length) ? (
+    <span>Loading....</span>
+  ) : (
+    <Table
+      headings={[
+        {
+          display: 'Name',
+          key: 'display_name',
+        },
+        {
+          display: 'Type',
+          key: 'content_type',
+        },
+        {
+          display: 'Date Added',
+          key: 'date_added',
+        },
+      ]}
+      data={assetsList}
+    />
+  )
+);
+
+AssetsTable.propTypes = {
+  assetsList: PropTypes.arrayOf(PropTypes.object).isRequired,
+};
+
+export default AssetsTable;
--- a/src/AssetsPage/index.jsx
+++ b/src/AssetsPage/index.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 import CheckBox from 'paragon/src/CheckBox';
-import Table from 'paragon/src/Table';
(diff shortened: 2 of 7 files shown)


src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Table from 'paragon/src/Table';
+
+const AssetsTable = ({ assetsList }) => (
```
arizzitano: FYI: this is a [stateless component](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-stateless-function.md). When a component has no methods other than `render`, it's recommended to rewrite it as stateless for simplicity's sake.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -0,0 +1,33 @@
...
+          key: 'display_name',
+        },
+        {
+          display: 'Type',
+          key: 'content_type',
+        },
+        {
+          display: 'Date Added',
+          key: 'date_added',
+        },
```
arizzitano: Just picked these headings randomly, but we can add more as needed.

src/AssetsPage/index.jsx
```diff
@@ -11,6 +15,10 @@ class AssetsPage extends React.Component {
     };
   }
 
+  componentDidMount() {
+    this.props.getAssets('course-v1:edX+DemoX+Demo_Course');
```
arizzitano: We should start thinking about how we're going to get the course id into React/Redux land. Presumably we'll pass it in from Studio when we instantiate the `AssetsPage` component?

src/api/client.js
```diff
@@ -12,7 +12,7 @@ export function pingStudioHome() {
   );
 }
 
-export function getAssets(courseId, { page = 0, pageSize = 50, sort = 'sort', assetType = '' }) {
+export function fetchAssets(courseId, { page = 0, pageSize = 50, sort = 'sort', assetType = '' }) {
```
MichaelRoytman: Could you clarify the diffence between getAssets and fetchAssets? They sound pretty similar in their intent. 

Is fetchAssets the action creator and getAssets the reducer? Could there be a way to make it more clear the difference between the two?

arizzitano: `fetchAssets` is the method in the API client that actually performs the low level `fetch` against the API; `getAssets` is the action creator itself which fires `fetchAssets` and then handles its response. I named this method `fetchAssets` so it wouldn't get confused with `getAssets` (and also so there wouldn't be a naming collision when I imported it within `src/data/actions/assets.js`).

I agree that these names aren't great, though. How's this solution?
* rename `getAssets` to `requestAssets`
* rename `assetsResponse` to `requestAssetsSuccess`
* rename `ASSETS_RESPONSE` to `REQUEST_ASSETS_SUCCESS`
* rename `fetchAssets` back to `getAssets`

As we scale up the number of XHRs we make, we will need to think about genericizing some of this machinery, but that'll need to be tackled in a followon PR.

src/AssetsPage/index.jsx
efischer19: That seems like a reasonable assumption to make at this stage.

src/api/client.js
efischer19: Whatever we end up deciding (I personally don't have a strong preference, @arizzitano's last suggestion seems good), this definitely needs to be documented and consistent across all the components we end up building.

arizzitano: Let's discuss in the arch meeting today!

PR #25, opened 2017-09-18
Asset type filter component

thallada:
I went over most of this in the Arch meeting today, but now the filter actually filters.

screenshot-20170918161007-1738x735

Example redux state with "Images" checked:

screenshot-20170918161123-476x227

@edx/educator-dahlia


src/api/client.js
```diff
@@ -12,7 +12,20 @@ export function pingStudioHome() {
   );
 }
 
-export function getAssets(courseId, { page = 0, pageSize = 50, sort = 'sort', assetType = '' }) {
+function assetTypeValueFromState(assetFilters) {
+  // TODO: change this to multi-value once API supports it
```
thallada: @MichaelRoytman once your API PR merges, this function can be updated to output a comma-separated string instead of just one asset_type.

src/api/client.js
```diff
@@ -12,7 +12,20 @@ export function pingStudioHome() {
...
+  if (assetFilters.other) {
+    return 'OTHER';
+  } else if (assetFilters.images) {
+    return 'Images';
+  } else if (assetFilters.documents) {
+    return 'Documents';
+  }
```
arizzitano: Scratchpad idea around simplifying some of this logic: what if we stored these param names (`OTHER`, `Images`, `Documents`) as the `key`s in `ASSET_TYPES` up above? The reducer state might get a little uglier, since we'd be using stuff like `OTHER` as keys there too, but we wouldn't have to deal with disambiguation around relating arbitrary `key`s to API param names.

src/data/reducers/assetsFilters.js
```diff
@@ -0,0 +1,20 @@
+import { FILTER_UPDATED } from '../actions/assets';
+
+const initialState = {
+  courseId: 'course-v1:edX+DemoX+Demo_Course',
```
arizzitano: Probably don't want to store this `courseId` as a default forever. This'll serve our purposes for now -- in the future, we'll need some way to get `courseId` from Django-land into Redux-land, so we don't have to worry about hardcoding anything.

src/data/reducers/assetsFilters.js
```diff
@@ -0,0 +1,20 @@
...
+  page: 0,
+  images: false,
+  documents: false,
+  other: false,
```
arizzitano: Because the URI doesn't have any filter params applied by default, you *could* get away with just not storing the `false` defaults at all -- one less place we'd have to hardcode these filter keys. Since these filter-specific parameters (`images`, `documents`, `other`) are closely related to each other, but not as related to `page` and `courseId`, I'd suggest nesting them into one arbitrary object in this reducer (e.g. `filters: {}`). Side benefit: when it's time to assemble the URI with multiple filter params in the api client, we can do something slick like
```
Object.keys(assetFilters.filters).reduce((memo, key) => {
    assetFilters.filters[key] && memo.push(assetFilters.filters[key]);
    return memo;
}).join(',');
```
@MichaelRoytman and I messed around with something similar last week.

src/data/actions/assets.js
```diff
@@ -1,16 +1,23 @@
...
+    getAssets(assetsFilters.courseId, {
+      page: assetsFilters.page,
+      assetsFilters,
     })
       .then(response => response.json())
       .then(json => dispatch(requestAssetsSuccess(json)));
+
+export const filterUpdate = (filterKey, filterValue) => ({
+  type: FILTER_UPDATED,
+  data: { [filterKey]: filterValue },
```
arizzitano: Love the programmatic key assignment syntax, def an underrated ES2015 feature =)

src/AssetsPage/index.jsx
```diff
@@ -45,12 +48,16 @@ class AssetsPage extends React.Component {
 
 AssetsPage.propTypes = {
   assetsList: PropTypes.arrayOf(PropTypes.object).isRequired,
+  assetsFilters: PropTypes.objectOf(
+    PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.bool]),
```
arizzitano: Good use of `PropTypes`!

src/data/reducers/assetsFilters.js
```diff
@@ -0,0 +1,20 @@
...
+};
+
+const assetsFilters = (state = initialState, action) => {
```
arizzitano: Could we give this guy a more generic name than `assetsFilters`? `images`, `documents`, and `other` are definitely filters, but `courseId` and `page` less so. Maybe something like `assetsParameters`?

src/data/reducers/assetsFilters.js
thallada: How about:

```javascript
const initialState = {
  courseId: 'course-v1:edX+DemoX+Demo_Course',
  page: 0,
  pageSize: 50,
  assetTypes: {},
  sort: 'sort',
};
```
If we ever add another filter set to the table, we could just add another key and object to this structure.

arizzitano: LGTM!

PR #28, opened 2017-09-21
Add devstack install make target

efischer19:
This seems to be working locally.

Test steps: first, install devstack as usual:
```
git clone git@github.com:edx/devstack.git
cd devstack
make requirements
make dev.clone
make provision
<wait a while>
make dev.up.all
```

Then, install a local development version of studio-frontend
```
<assumes you're still in the devstack dir from above>
cd ../src
git clone git@github.com:edx/studio-frontend.git
cd studio-frontend
make devstack-install
```

After doing this, I saw 
```
CHANGED: node_modules/@edx/studio-frontend/src/BackendStatusBanner/styles.scss
---> pavelib.assets.compile_sass
```
in my lms_watcher logs, so I think it went through correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,3 +38,21 @@ from-scratch: ## start development environment from scratch
 	docker rmi edxops/studio-frontend
 	docker build -t edxops/studio-frontend:latest --no-cache .
 	make up
+
+devstack-install: ## install local version of package into docker devstack for development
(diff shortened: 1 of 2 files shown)


efischer19:
I have yet to add the "publish a new version to npm" make target we discussed as standup, that'll be the next addition.

PR #31, opened 2017-09-28
make devstack-install

efischer19:
WIP

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,11 +39,27 @@ from-scratch: ## start development environment from scratch
 	docker build -t edxops/studio-frontend:latest --no-cache .
 	make up
 
-devstack-install: ## install local version of package into docker devstack for development
-	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/nodeenvs/edxapp/bin/activate && cd /edx/app/edxapp/edx-platform && npm uninstall @edx/studio-frontend && cd /edx/src/studio-frontend && npm link && cd /edx/app/edxapp/edx-platform && npm link @edx/studio-frontend'
+devstack-install: validate-devstack-folders ## install local version of package into docker devstack for development
+	# note that this assumes a structure /src/sfe_parent/studio-frontend
+	# step 0: remove studio-frontend from edx-platform/node_modules
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/nodeenvs/edxapp/bin/activate && cd /edx/app/edxapp/edx-platform && npm uninstall @edx/studio-frontend && rm -rf ./node_modules/@edx/studio-frontend'
+	# step 1: create symlink in /edx/src/sfe_parent/symlinks
+	# note that we delete studio-frontend/node_modules; this is deliberate
+	docker exec -t edx.devstack.studio bash -c 'mkdir -p /edx/src/sfe_parent/symlinks && rm -rf /edx/src/sfe_parent/symlinks/* && source /edx/app/edxapp/edxapp_env && cd /edx/src/sfe_parent/studio-frontend && export npm_config_prefix=/edx/src/sfe_parent/symlinks && npm link && rm -rf /edx/src/sfe_parent/studio-frontend/node_modules'
+	# step 2: install symlinked package under edx-platform
+	# if studio-frontend/node_modules exists during this step, it can cause issues
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && export npm_config_prefix=/edx/src/sfe_parent/symlinks && npm link @edx/studio-frontend'
+	# step 3: repair installations
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && npm update && npm install'
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend && npm install && npm run build'
+	# step 4: update assets properly, so webservers know about them
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_assets'
+
+devstack.update: ## use this if you don't want to fire up the dedicated asset watching containers
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver webpack'
 
 asset-page-flag: ## insert a waffle flag into local docker devstack
-	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && echo "from cms.djangoapps.contentstore.config.models import NewAssetsPageFlag; NewAssetsPageFlag.objects.all().delete(); NewAssetsPageFlag.objects.create(enabled=True, enabled_for_all_courses=True);" | ./manage.py lms --settings=devstack_docker shell && echo "NewAssetsPageFlag inserted!"'
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && echo "from cms.djangoapps.contentstore.config.models import NewAssetsPageFlag; NewAssetsPageFlag.objects.all().delete(); NewAssetsPageFlag.objects.create(enabled=True, enabled_for_all_courses=True);" | ./manage.py lms --settings=devstack_docker shell && echo "NewAssetsPageFlag inserted!"'
 
 publish: publish-patch ## default is a path release
 
@@ -52,3 +68,6 @@ publish-%: ## publish a new version from master. Argument will be fed into `npm 
 	@git diff --quiet || (echo 'unclean git repo, please commit all changes before publish'; exit 1)
 	export VERSION=$$(npm version patch) && git checkout -b dahlia/$$VERSION && npm publish --access public && git push --set-upstream origin dahlia/$$VERSION && git push --tags
 	echo "NPM package published, git branch created and tags pushed. Go to https://github.com/edx/studio-frontend to see your PR and merge the package.json update"
+
+validate-devstack-folders:
+	export TEMP_FILE=$$(date +%s) && touch $$TEMP_FILE && docker exec -t edx.devstack.studio bash -c 'if [[ $$(ls /edx/src/sfe_parent/studio-frontend/$$TEMP_FILE) ]]; then echo "Folder structure looks good"; else echo "Your folders are FUBAR, talk to Eric (who should really come up with an automatic way of repairing this situation)"; exit 1; fi' && rm -rf $$TEMP_FILE
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   "dependencies": {
     "@edx/paragon": "^0.0.1",
     "classnames": "^2.2.5",
+    "@edx/edx-bootstrap": "0.3.3",
     "font-awesome": "^4.7.0",
     "popper.js": "^1.12.5",
     "prop-types": "^15.5.10",


Makefile
```diff
@@ -40,7 +40,19 @@ from-scratch: ## start development environment from scratch
...
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/nodeenvs/edxapp/bin/activate && cd /edx/app/edxapp/edx-platform && npm uninstall @edx/studio-frontend && rm -rf ./node_modules/@edx/studio-frontend'
+	# step 1: create symlink in /edx/src/studio-frontend/symlinks
+	# note that we delete studio-frontend/node_modules; this is deliberate
+	docker exec -t edx.devstack.lms bash -c 'mkdir -p /edx/src/studio-frontend/symlinks && rm -rf /edx/src/studio-frontend/symlinks/* && source /edx/app/edxapp/edxapp_env && cd /edx/src/studio-frontend && export npm_config_prefix=/edx/src/studio-frontend/symlinks && npm link && rm -rf /edx/src/studio-frontend/node_modules'
+	# step 2: install symlinked package under edx-platform
+	# if studio-frontend/node_modules exists during this step, it can cause issues
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && export npm_config_prefix=/edx/src/studio-frontend/symlinks && npm link @edx/studio-frontend'
+	# step 3: repair installations
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && npm update && npm install'
```
efischer19: Step 3.1:

* In edx-platform, `npm update && npm install`. Since none of studio-frontend's dependencies live under `node_modules/studio-frontend/node_modules`, the dependencies are freshly fetched from the internet

Makefile
```diff
@@ -40,7 +40,19 @@ from-scratch: ## start development environment from scratch
...
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend && npm install && npm run build'
```
efischer19: Step 3.2:

* In studio-frontend (since the symlink is in place, we could do this in either `/edx/src/studio-frontend` *or* `/edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend`), `npm install && npm run build`
* This "fixes" studio-frontend's installation (I forget exactly how, but at one point it broke without the 'install')
* Also, since edx-platform doesn't specify prod-versus-dev configs, we need to build `dist/`, as the CMS expects it to be there.

Makefile
```diff
@@ -40,7 +40,19 @@ from-scratch: ## start development environment from scratch
...
+	# step 4: update assets properly, so webservers know about them
+	docker exec -t edx.devstack.lms bash -c 'paver update_assets'
```
efischer19: Finally, `update_assets`. This will have the effect of:

* running webpack
* collecting generated assets and placing them in a place where the CMS may find them

package.json
```diff
@@ -11,7 +11,10 @@
   },
   "license": "AGPL-3.0",
   "dependencies": {
+    "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-object-rest-spread": "^6.26.0",
     "classnames": "^2.2.5",
+    "edx-bootstrap": "^0.2.1",
```
efischer19: I had to move all 3 of these into `dependencies` so that edx-platform would install them, but I'm not 100% sure I remember why.

efischer19:
Pairs with https://github.com/edx/devstack/pull/240 and https://github.com/edx/edx-platform/pull/16111

Makefile
```diff
@@ -40,7 +40,23 @@ from-scratch: ## start development environment from scratch
...
+	# step 3: repair installations
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && npm update && npm install'
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend && npm install && npm run build'
+	# step 4: update assets properly, so webservers know about them
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_assets'
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_assets'
+
+devstack.update: ## moves newly webpack-generated assets to a place where studio can find them
+	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver webpack && python ./manage.py cms --settings=devstack_docker collectstatic --noinput'
```
efischer19: This took ~1:20 on my machine, and resulted in updated studio-frontend assets

Makefile
```diff
@@ -40,10 +40,26 @@ from-scratch: ## start development environment from scratch
...
+	# if studio-frontend/node_modules exists during this step, it can cause issues
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && export npm_config_prefix=/edx/src/sfe_parent/symlinks && npm link @edx/studio-frontend'
+	# step 3: repair installations
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && npm update && npm install'
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend && npm install && npm run build'
+	# step 4: update assets properly, so webservers know about them
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver update_assets'
+
+devstack.update: ## moves newly webpack-generated assets to a place where studio can find them
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver webpack && python ./manage.py cms --settings=devstack_docker collectstatic --noinput'
```
efischer19: This won't be needed if the asset watchers are running; do we want to leave it in anyway for manual updates?

efischer19: Also I need to test this - I don't believe the final `collectstatic` is needed any longer.

package.json
efischer19: I think it had to do with the loaders I places in edx-platform's webpack config

efischer19:
Everything still works as described yesterday; the key difference is that now our folder structure looks like this:
```
+ src/
|
+--+ studio_frontend/
|
+--+ symlinks/
```
instead of this:
```
+ src/
|
+--+ studio_frontend/
   |
   + symlinks/
```


Thus, the asset watchers can watch `edx-platform/node_modules/@edx/studio-frontend` without falling into an infinite directory loop.

Makefile
efischer19: Confirmed - when the asset watchers are *not* running (`dev.up` rather than `dev.up.all`), the initial `source /edx/app/edxapp/edxapp_env && cd /edx/app/edxapp/edx-platform && paver webpack` is all that's needed to force a cms update of your locally-installed `studio-frontend` code.

efischer19:
I've cleaned this up to include the results of my `devstack.update` test, and rebased to fix the merge issue github was complaining about.

Makefile
```diff
@@ -40,10 +40,26 @@ from-scratch: ## start development environment from scratch
...
 
 devstack-install: ## install local version of package into docker devstack for development
-	docker exec -t edx.devstack.lms bash -c 'source /edx/app/edxapp/nodeenvs/edxapp/bin/activate && cd /edx/app/edxapp/edx-platform && npm uninstall @edx/studio-frontend && cd /edx/src/studio-frontend && npm link && cd /edx/app/edxapp/edx-platform && npm link @edx/studio-frontend'
+	# note that this assumes a structure /src/sfe_parent/studio-frontend
+	# step 0: remove studio-frontend from edx-platform/node_modules
+	docker exec -t edx.devstack.studio bash -c 'source /edx/app/edxapp/nodeenvs/edxapp/bin/activate && cd /edx/app/edxapp/edx-platform && npm uninstall @edx/studio-frontend && rm -rf ./node_modules/@edx/studio-frontend'
+	# step 1: create symlink in /edx/src/sfe_parent/symlinks
+	# note that we delete studio-frontend/node_modules; this is deliberate
+	docker exec -t edx.devstack.studio bash -c 'mkdir -p /edx/src/sfe_parent/symlinks && rm -rf /edx/src/sfe_parent/symlinks/* && source /edx/app/edxapp/edxapp_env && cd /edx/src/sfe_parent/studio-frontend && export npm_config_prefix=/edx/src/sfe_parent/symlinks && npm link && rm -rf /edx/src/sfe_parent/studio-frontend/node_modules'
```
MichaelRoytman: Does npm link use the npm_config_prefix?

Makefile
MichaelRoytman: Echoing Farhanah's comment this morning, did you get a chance to test this? What were the outcomes?

Makefile
efischer19: Yes - [link 1](https://docs.npmjs.com/cli/link#description), [link 2](https://docs.npmjs.com/misc/config#prefix)

```
First, npm link in a package folder will create a symlink in the global folder {prefix}/lib/node_modules/<package> that links to the package where the npm link command was executed. (see npm-config for the value of prefix). It will also link any bins in the package to {prefix}/bin/{name}.
```

MichaelRoytman: Thanks!

I could very well be misunderstanding something, but where does the studio-frontend subdirectory of sfe_parent get created? I see that you note that it assumes the structure /src/sfe_parent/studio-frontend, but where is that created?  Is that something we need to enforce in order for this to work? I couldn't find any other reference to it elsewhere, so I assume you'd have to manually create it before running this command? Could you clarify?

efischer19: > so I assume you'd have to manually create it before running this command

This is how I "solved" the chicken/egg problem of needing to describe the folder structure that contains studio-frontend in a file inside studio-frontend. I usually create directories with git clone, like this:

```
(after starting devstack containers)
ericfischer@Eric-Fischers-MacBook-Pro:~/work/devstack$ cd ../src/
ericfischer@Eric-Fischers-MacBook-Pro:~/work/src$ mkdir sfe_parent
ericfischer@Eric-Fischers-MacBook-Pro:~/work/src$ cd sfe_parent
ericfischer@Eric-Fischers-MacBook-Pro:~/work/src/sfe_parent$ git clone git@github.com:edx/studio-frontend.git
...
ericfischer@Eric-Fischers-MacBook-Pro:~/work/src/sfe_parent$ cd studio-frontend
ericfischer@Eric-Fischers-MacBook-Pro:~/work/src/sfe_parent/studio-frontend$ git checkout efischer/dependency_funtimes
...
ericfischer@Eric-Fischers-MacBook-Pro:~/work/src/sfe_parent/studio-frontend$ make devstack-install
```

Is there any easy way (in Makefile) to check the structure and error out if it's not correct? I'll look into that further

MichaelRoytman: Ah, I see. Could this be useful (second answer) https://stackoverflow.com/questions/5553352/how-do-i-check-if-file-exists-in-makefile? The wildcard could be promising.

package.json
efischer19: Resolved thanks to @ari - we ought to move these deps into edx-platform, since its webpack process is the one that needs them.

arizzitano: xposted from HipChat, preserving here from posterity: I'd recommend keeping these as devDependencies in SFE, but adding them as dependencies in edx-platform. Since edx-platform essentially owns its own webpack process, it makes sense for it to have explicit dependencies for packages its webpack process consumes. Normally we'd add these as devDependencies in edx-platform as well, but due to the way we build and deploy assets, anything that's required for a production build needs to go in `dependencies`.

package.json
```diff
@@ -32,6 +32,7 @@
     "babel-preset-env": "^1.6.0",
     "babel-preset-react": "^6.24.1",
     "css-loader": "^0.28.5",
+    "edx-bootstrap": "^0.2.1",
```
efischer19: I believe this new devDependency is required due to https://github.com/edx/studio-frontend/blob/36782a58ade9e115a1b3231fc45efd67ef4ce685/config/webpack.dev.config.js#L54, does that seem right?

arizzitano: Oooh, good catch -- that should probably actually go in prod `dependencies`. I'm surprised we weren't getting errors about that!

efischer19: Oh yeah, I didn't notice it in `webpack.prod.config.js` the first time. Updating...

arizzitano: Actually, it'd be good to switch versions for this package since Andy recently changed its name. It is now `"@edx/edx-bootstrap": "0.3.3"`

efischer19: Is that going to land in edx-platform soon as well?

https://github.com/edx/edx-platform/blob/master/package.json#L14

arizzitano: It probably should. @andy-armstrong -- are we good to add the new `@edx/edx-bootstrap` dependency into platform?

andy-armstrong: @arizzitano I made that change in my hackathon PR but that is on hold temporarily. If you guys want to go ahead and make that change that would be great, or I can do it.

efischer19: I'll add it to https://github.com/edx/edx-platform/pull/16111, thanks @andy-armstrong!

Makefile
MichaelRoytman: Were you able to look into whether it would be worthwhile to verify the directory structure is correct to avoid issues? Or should we add a make target somewhere to do this work for the user?

efischer19: Ah yes, I had forgotten about that :/

I'll get something up to address it.

Makefile
```diff
@@ -52,3 +68,6 @@ publish-%: ## publish a new version from master. Argument will be fed into `npm
 	@git diff --quiet || (echo 'unclean git repo, please commit all changes before publish'; exit 1)
 	export VERSION=$$(npm version patch) && git checkout -b dahlia/$$VERSION && npm publish --access public && git push --set-upstream origin dahlia/$$VERSION && git push --tags
 	echo "NPM package published, git branch created and tags pushed. Go to https://github.com/edx/studio-frontend to see your PR and merge the package.json update"
+
+validate-devstack-folders:
+	docker exec -t edx.devstack.studio bash -c 'if [[ $$(ls /edx/src/sfe_parent/symlinks) && $$(ls /edx/src/sfe_parent/studio-frontend) ]]; then echo "Folder structure looks good"; else echo "Your folders are FUBAR, talk to Eric (who should really come up with an automatic way of repairing this situation)"; exit 1; fi'
```
efischer19: This verifies that the folders exist on the devstack containers. Is that good enough, or should we add in tougher checks like "the reported git status of the container folder matches the folder we're currently executing from"?

PR #32, opened 2017-09-29
Add FA trashcan icon to AssetsTable

fysheets:
This adds a button to the UI using the FontAwesome trashcan icon as a part of the Files & Uploads table that will enable users to delete the asset that it is correlated to.

### DIFF
--- a/src/AssetsPage/AssetsTable/index.jsx
+++ b/src/AssetsPage/AssetsTable/index.jsx
@@ -26,7 +26,11 @@ class AssetsTable extends React.Component {
   addDeleteButton() {
     const newAssetsList = this.props.assetsList.map((asset) => {
       const currentAsset = Object.assign({}, asset);
-      const deleteButton = (<Button className={['fa', 'fa-trash', styles['icon']]} display='' buttonType='link' />);
+      const deleteButton = (<Button
+        className={['fa', 'fa-trash', styles.icon]}
+        display={''}
+        buttonType={'link'}
+      />);
       currentAsset.delete_asset = deleteButton;
       return currentAsset;
     });
--- a/src/AssetsPage/AssetsTable/index.jsx
+++ b/src/AssetsPage/AssetsTable/index.jsx
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import Table from 'paragon/src/Table';
 import Button from 'paragon/src/Button';
 
-import styles from '../styles.scss';
+import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
 
 class AssetsTable extends React.Component {
   constructor(props) {
@@ -11,6 +11,9 @@ class AssetsTable extends React.Component {
     this.state = {
       finalAssetsList: [],
     };
+
+    this.onDeleteClick = this.onDeleteClick.bind(this);
+    this.addDeleteButton = this.addDeleteButton.bind(this);
   }
 
   componentDidMount() {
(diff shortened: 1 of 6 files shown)


src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -1,30 +1,85 @@
...
+    super(props);
+    this.state = {
+      finalAssetsList: [],
+    };
+
+    this.onDeleteClick = this.onDeleteClick.bind(this);
+    this.addDeleteButton = this.addDeleteButton.bind(this);
+  }
+
+  componentDidMount() {
```
arizzitano: Do we need this blank method?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -1,30 +1,85 @@
...
+  }
+
+  addDeleteButton() {
+    const newAssetsList = this.props.assetsList.map((asset) => {
+      const currentAsset = Object.assign({}, asset);
+      const deleteButton = (<Button
+        className={[FontAwesomeStyles.fa, FontAwesomeStyles['fa-trash']]}
+        display={''}
+        buttonType={'light'}
+        aria-label={`Delete current asset named ${currentAsset.display_name}`}
```
arizzitano: I might actually opt towards making this a little less verbose (e.g. `Delete asset ${currentAsset.display_name}`), but I'm curious what @cptvitamin's thoughts are.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -1,30 +1,85 @@
...
-    />
-  )
-);
+import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
+
+class AssetsTable extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      finalAssetsList: [],
```
arizzitano: Just a nit: I might name this something like `displayAssetsList` -- it's not exactly *final* since we update it when the underlying data changes ` -- but up to you.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -1,30 +1,85 @@
...
+
+  onDeleteClick() {
+    this.setState({
+      // TODO: Update application state for Modal and delete logic
+    });
+  }
+
+  addDeleteButton() {
+    const newAssetsList = this.props.assetsList.map((asset) => {
+      const currentAsset = Object.assign({}, asset);
```
arizzitano: :+1: for creating a fresh list of assets!

src/AssetsPage/AssetsTable/index.jsx
arizzitano: Hmm actually -- do you think it would make sense to fire `addDeleteButton` in here too? Just in case the table gets initialized with all the props it needs and doesn't fire any `componentDidUpdate`s.

src/AssetsPage/AssetsTable/index.jsx
cptvitamin: I agree with @arizzitano.  I'm a fan of succinct messaging for everyone.  However, clear, concise instructions are of *paramount* importance for screen reader users who are already exposed to so much verbosity/chatter due to the very nature with which they interact with applications.  I'm not even sure its necessary to include the word "asset" here, since that may already be clear from the context.  Let's go with ```aria-label={`Delete  ${currentAsset.display_name}`}```

src/AssetsPage/AssetsTable/index.jsx
fysheets: I think that was the initial use of this method before I moved it to the `componentDidUpdate` - I can add another call into this method to catch both on initial load & if the list updates.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -39,7 +39,7 @@ class AssetsTable extends React.Component {
         className={[FontAwesomeStyles.fa, FontAwesomeStyles['fa-trash']]}
         display={''}
         buttonType={'light'}
-        aria-label={`Delete current asset named ${currentAsset.display_name}`}
+        aria-label={`Delete ${currentAsset.display_name}`}
```
cptvitamin: 👍

arizzitano:
LGTM! 🚢

PR #33, opened 2017-10-10
mroytman/integrate modal

MichaelRoytman:


.eslintrc.json
```diff
@@ -9,6 +9,9 @@
           "config/*.js"
         ]
       }
+    ],
+    "class-methods-use-this": [
```
MichaelRoytman: ESLint was complaining about the renderBody function of AssetsPage/AssetsTable/index.jsx, so we've chosen to ignore it here, as we find it to be an unhelpful case.

config/webpack.dev.config.js
```diff
@@ -77,7 +77,9 @@ module.exports = Merge.smart(commonConfig, {
     port: 18011,
     proxy: Object.keys(apiEndpoints).reduce(
       (map, endpoint) => {
-        map[apiEndpoints[endpoint]] = targetUrl;// eslint-disable-line no-param-reassign
+        map[apiEndpoints[endpoint]] = {
+          target: targetUrl,
+        };
```
MichaelRoytman: I've changed this reduce function to return an object of the form:

proxy: {
    "aaa": {...},
    "bbb": {...},
}

as I think it will be easier to add to, if necessary. I don't think it's necessary, but thought I'd include it to field thoughts on it.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -52,37 +65,97 @@ class AssetsTable extends React.Component {
...
 AssetsTable.propTypes = {
   assetsList: PropTypes.arrayOf(PropTypes.object).isRequired,
+  assetsParameters: PropTypes.objectOf(
+    PropTypes.oneOfType([PropTypes.string, PropTypes.number, PropTypes.bool, PropTypes.object]),
+  ).isRequired,
+  deleteAsset: PropTypes.func.isRequired,
 };
 
-export default AssetsTable;
+const mapStateToProps = state => {
```
MichaelRoytman: I pulled out the mapStateToProps and mapDispatchToProps functions from the connect function call, since I thought it was a little more clear what the connect function was doing this way. When I was looking at WrappedAssetsFilters as an example, it took me some time to realize what the connect function was doing, so I think that giving the functions explicit "Google-able" names may be helpful for people unfamiliar with react-redux, instead of them being defined inside the connect call. Any thoughts?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -52,37 +65,97 @@ class AssetsTable extends React.Component {
...
+        />
+        {this.renderModal()}
+      </div>
     );
   }
 }
 
 AssetsTable.propTypes = {
   assetsList: PropTypes.arrayOf(PropTypes.object).isRequired,
+  assetsParameters: PropTypes.objectOf(
```
MichaelRoytman: assetsParameters is necessary because it's needed by the deleteAsset action creator (it contains the courseID of the course in which the asset exists), which is dispatched from the deleteAsset function. I had to add it to the AssetTable to be able to dispatch the action creator from deleteAsset. Is there a better way? It seemed a little out of place here.

src/data/actions/assets.js
```diff
@@ -1,7 +1,10 @@
-import { getAssets } from '../../api/client';
+import * as clientApi from '../../api/client';
```
MichaelRoytman: I thought this was a little more clear to know where the getAssets function is coming from by referring to the clientApi explicitly, but I don't feel strongly about it. It was also to allow for me to use the deleteAsset function name in this file and in the client api without name conflicts.

src/data/actions/assets.js
```diff
@@ -21,3 +24,25 @@ export const filterUpdate = (filterKey, filterValue) => ({
...
+});
+
+export const assetXHRFailure = (text) => ({
+  type: ASSET_XHR_FAILURE,
+  text,
+});
+
+export const deleteAsset = (assetsParameters, assetId) =>
+  dispatch =>
+    clientApi.deleteAsset(assetsParameters.courseId, assetId)
```
MichaelRoytman: I hesitate a little to have this function be called deleteAsset and the client API function to be deleteAsset, since they do semantically different things: one is an action creator and the other is actually making an API request. I'd like for the names to be differently, but I can't think of a not-clunky solution. deleteAsset here and requestAssetDelete in the API? Any thoughts? It's not a big deal, but I figure I'd rather resolve it now rather than later.

src/api/client.js
```diff
@@ -28,3 +29,16 @@ export function getAssets(courseId, { page = 0, pageSize = 50, sort = 'sort', as
...
 }
+
+export function deleteAsset(courseId, assetId) {
+  return fetch(
+    `${endpoints.assets}/${courseId}/${assetId}`, {
+      credentials: 'same-origin',
+      method: 'delete',
+      headers: {
+        Accept: 'application/json',
+        'X-CSRFToken': Cookies.get('csrftoken'),
```
MichaelRoytman: TODO: Ari suggested having a top-level request object that has common attributes that other requests can extend, so we don't have to add common headers to each request.

package.json
```diff
@@ -13,7 +13,8 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "font-awesome": "^4.7.0",
-    "paragon": "edx/paragon#bdde7afe5d1ade1dce83f50112e778b6d6d25215",
+    "js-cookie": "^2.1.4",
+    "paragon": "edx/paragon#fd4b1bdd6690bf97b7ed57a0d209faae96095081",
```
MichaelRoytman: This is so I can use the modal.

src/api/client.js
MichaelRoytman: I have added it to my fancy TODO chart.

src/data/reducers/assets.js
```diff
@@ -0,0 +1,47 @@
+import { REQUEST_ASSETS_SUCCESS, DELETE_ASSET_SUCCESS, ASSET_XHR_FAILURE, FILTER_UPDATED } from '../actions/assets';
```
MichaelRoytman: Is there a better way to do this? This is not gonna be very pretty as we have more actions.

fysheets: You can do something like `import * as AssetsActions from '../actions/assets';` -- then when referencing below do something like `AssetsActions.DELETE_ASSET_SUCCESS`

src/data/actions/assets.js
fysheets: What about `deleteAssetApi`? Since it's an API call? We can talk through this some more if you'd like too.

src/AssetsPage/AssetsTable/index.jsx
fysheets: If you're still wondering about this one, I can take a look at how this is set up more today, and see if I can remember if there's another way to handle it.

MichaelRoytman:
TODO: I need to address focus on the assets page. For some reason, the focus does not display in a perceivable way on the trashcan button and on the x button within the modal, so it's hard to visually verify how the focus behaves. Any idea why?

I need to add logic for focus changes in the following ways:
1. If a modal is closed without deleting the asset, the focus should return to the trashcan icon that was clicked to open the modal.
2. If a modal is closed by pressing the delete message, the focus should move to the status message.

I'm not sure if we have agreed on 2 or not.

src/data/reducers/assets.js
MichaelRoytman: I think @fysheets was going to propose moving the action objects out of actions/assets.js into something like actions/assetsActionTypes.js and actions creators into assetsActionCreators.js. If we move in that direction, we could do import * without importing the action creators too. Would it make sense to do that, assuming that recommendation is how we want to move forward? I don't think it would be a big pain to switch between the two if they're being used the same way in the assets reducer for the time being.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -2,18 +2,30 @@ import React from 'react';
...
+
 
 class AssetsTable extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
       displayAssetsList: [],
+      modalOpen: false,
+      assetToDelete: {},
+      elementToFocusOnModalClose: {},
```
MichaelRoytman: Any better ideas for this name? I originally had 'focusableElement', but I don't think that it was clear that it was only being used for focus upon closing the modal.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -2,18 +2,30 @@ import React from 'react';
...
     };
 
     this.onDeleteClick = this.onDeleteClick.bind(this);
+    this.deleteAsset = this.deleteAsset.bind(this);
     this.addDeleteButton = this.addDeleteButton.bind(this);
+    this.closeModal = this.closeModal.bind(this);
+
+    this.trashcanRefs = {};
```
MichaelRoytman: I need to store the references to the trashcan icons in order to handle focus. The object is keyed by asset.id, and the value is the reference to the icon.

MichaelRoytman: Since focus isn't being highlighted in the browser ideally, you can use this event listener to see which elements have focus after a keystroke.

document.addEventListener("keyup", function() {
    console.log(document.activeElement);
});

Just enter that in the developer console, and it should log which button has focus as you're tabbing through.

arizzitano: :+1: good tip! I'll look into why the focus outline isn't showing - my suspicion is it has something to do with the way we're scoping Bootstrap styles.

arizzitano:
@MichaelRoytman looks great! Last request, can you bump the version (0.2.0) and then cut a new release after merging?

PR #35, opened 2017-10-11
mroytman/testing infrastructure

MichaelRoytman:


.travis.yml
```diff
@@ -5,10 +5,14 @@ node_js:
 
 sudo: required
 services:
-  - docker
+   - docker
 
 before_script:
-  - docker build -t studio-frontend --no-cache .
+   - docker build -t studio-frontend --no-cache .
```
MichaelRoytman: Fix this.

.travis.yml
```diff
@@ -5,10 +5,14 @@ node_js:
 
 sudo: required
 services:
-  - docker
+   - docker
```
MichaelRoytman: Fix this.

.eslintrc.json
```diff
@@ -13,5 +18,8 @@
     "class-methods-use-this": [
```
arizzitano: I noticed the tests are throwing a `import/no-extraneous-dependencies` error -- fortunately this is easy to fix, just add `"*.test.jsx"` inside the `devDependencies` object up above (add it after line 14 -- i can't commend directly on that line for some reason). This tells eslint:
- for the `import/no-extraneous-dependencies` rule,
- if it applies to imports from `devDependencies`,
- skip it for all files that match the `*.test.jsx` pattern (because technically tests aren't part of production code, so we don't need to treat their deps as prod deps)

As you might notice, we're already applying this rule to everything within the config directory (which is also not *technically* part of production code) as well.

MichaelRoytman: I've added it. I needed to tack on a "**" to make it "**/*.test.jsx" in order for it to work. I'm not totally sure why, since the one above it worked fine, but maybe that's because it was specifying a top-level directory that has a flat structure inside instead of nested directories?

.eslintrc.json
```diff
@@ -1,17 +1,26 @@
 {
   "extends": "eslint-config-edx",
   "parser": "babel-eslint",
+  "parserOptions": {
```
MichaelRoytman: This is in eslint-config-edx, so we should remove this.

src/AssetsPage/index.jsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
-import AssetsTable from './AssetsTable';
+import { AssetsTable } from './AssetsTable';
```
fysheets: Is this change as well as the AssetsTable/index change to export the class going to be necessary for all future component classes?

MichaelRoytman: In order to write tests for the actual React component, you need to be able to import the non-connected component, because, otherwise, jest complains about not having a Redux store. So, instead of importing the WrappedAssetsTable, I need to import the AssetsTable; therefore, I have to do a named export of AssetsTable. Does that make sense? If we want to test our components independent of their Redux stores, then, yes, we'll have to do this for each component, as we write unit tests for them. I believe you can test Redux separately anyway, but I'll defer to Ari on the best way to position ourselves relative to testing going forward.

fysheets: That does make sense, thanks! Wanted to sanity check understanding on what steps would be needed for future testing.

PR #39, opened 2017-10-18
i18n proof-of-concept

efischer19:
This might work, let's find out


efischer19:
In action screenshot: screen shot 2017-10-18 at 4 34 13 pm

.babelrc
```diff
@@ -0,0 +1,30 @@
...
+  "env": {
+    "i18n": {
+      "plugins": [
+        ["react-intl", {
+          "messagesDir": "./i18n/default/",
+          "enforceDescriptions": true
+        }]
+      ]
+    }
+  }
```
efischer19: I'm not super familiar with babel, so there's a better-than-zero chance that this `.babelrc` could be written more cleanly/correctly.

i18n/default/src/AssetsPage/index.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "id": "AssetsPage.title",
+    "description": "Title for AssetsPage",
+    "defaultMessage": "Files & Uploads"
+  }
+]
```
efischer19: Note that the extracted messages are [MessageDescriptor objects](https://github.com/yahoo/react-intl/wiki/Components#message-descriptor)

i18n/locales/fr.json
```diff
@@ -0,0 +1,3 @@
+{
+  "AssetsPage.title": "Oui Oui, I totally know French"
+}
```
efischer19: But the translation strings are just `"id": "translated string"` pairs in JSON. Note that this is a totally fake translation, I don't actually know 
français. But I figure the transifex step shouldn't be a hugely difficult problem, yeah?

i18n/locales/whitelist_fr.json
```diff
@@ -0,0 +1,2 @@
+[
+]
```
efischer19: I'm not 100% on the usage of this, but it's a result of `react-intl-translations-manager`

package.json
```diff
@@ -43,30 +47,11 @@
...
-            ]
-          }
-        }
-      ],
-      "babel-preset-react"
-    ],
-    "plugins": [
-      "transform-object-rest-spread",
-      "transform-class-properties"
-    ]
```
efischer19: I think @MichaelRoytman is yanking out `.babelrc` in his work; I replayed here b/c all the tutorials assumed that setup.

scripts/translationRunner.js
```diff
@@ -0,0 +1,8 @@
+// translationRunner.js
+const manageTranslations = require('react-intl-translations-manager').default;
+
+manageTranslations({
+  messagesDirectory: 'i18n/default/',
+  translationsDirectory: 'i18n/locales',
+  languages: ['fr'], // any translation---don't include the default language
+});
```
efischer19: Copied from https://mikebridge.github.io/articles/typescript-i18n-react-intl/, who copied it from https://github.com/GertjanReynaert/react-intl-translations-manager

src/AssetsPage/index.jsx
```diff
@@ -36,7 +38,12 @@ class AssetsPage extends React.Component {
   render() {
     return (
       <div className={styles.assets}>
-        <h2>Files & Uploads</h2>
+        <FormattedMessage
+          tagName="h2"
+          id="AssetsPage.title"
+          defaultMessage="Files & Uploads"
+          description="Title for AssetsPage"
+        />
```
efischer19: This is how we'll declare internationalizable strings in code.

src/index.jsx
```diff
@@ -1,19 +1,31 @@
...
 
 import AssetsPage from './AssetsPage';
 import BackendStatusBanner from './BackendStatusBanner';
 import store from './data/store';
 
+import frenchMessages from '../i18n/locales/fr.json';
+
+// en is hardcoded in right now, eventually this will be a dynamic locale
+addLocaleData(enLocaleData);
+addLocaleData(frLocaleData);
```
efischer19: This is going to be the difficult portion to solve. When the locale is hardcoded, I can easily import the correct messages (line 12) and `addLocaleData` with the correct data (required by `IntlProvider`, something to do with knowing how to format dates I think?).

But how will we be able to do this on-the-fly? I'm very open to ideas here - keep in mind that we *only* want to import `*LocaleData` and messages for the current language, not all of them.

Could we give each locale a small script, located at `../i18n/locales/<locale>.js` to load localedata and messages? Then we can (in platform) give each locale a webpacked entry point, and dynamically include it into the template at render time...

src/index.jsx
```diff
@@ -1,19 +1,31 @@
...
+
 const App = () => (
-  <Provider store={store}>
-    <div>
-      <BackendStatusBanner />
-      <h1>studio-frontend</h1>
-      <AssetsPage />
-    </div>
-  </Provider>
+  <IntlProvider locale="fr" messages={frenchMessages}>
```
efischer19: I believe that if `messages` is empty, the `defaultMessage` is used for each `FormattedMessage`

efischer19:
FYI @edx/educator-dahlia

i18n/default/src/AssetsPage/index.json
dsjen: Is this thinking that this is the file that we feed up to transifex for translation? Something to think about is that we'll eventually have more pages than just this one, and ideally we wouldn't have to specify to transifex each new page to translate explicitly.

src/index.jsx
dsjen: Yes, very tricky!

Just to check my understanding, let me restate. If we include each and every local data that we have, our webpack-ed bundle would grow as we add languages, pages, and as translations come in. We'd rather have *only* the correct language JSON loaded. Therefore, we don't want the language bundled.

Because the platform manages the bundle, we'd need to find a way for the language data to be injected in. Could we attach it to the `root` in a `data-*` attribute and then load it here?

i18n/default/src/AssetsPage/index.json
efischer19: Yeah, the thinking here is that we can use "every json file under `i18n/default` " as our input to transifex. I've currently got babel configured to extract all relevant display strings under `src/` into `i18n/default/src/`, mirroring the directory structure.

If the tooling is set up correctly, we shouldn't ever need devs to touch files under `i18n/` explicitly - babel tools will extract inline strings from `src/` and place them under `i18n/default/src`, where we can easily feed them to transifex. Then transifex can output the json under `i18n/<locale>`, for all locales.

This works well for transifex, but as it's currently set up we're placing *all* translations for a given locale into a single file - should we consider keeping the transifex output stored per-page, to optimize bundle sizing?

efischer19:
This will never merge; but may still be useful to the actual i18n implementation work.

PR #40, opened 2017-10-18
Integrate StatusAlert into Delete Workflow

fysheets:
Integrate the proper Paragon StatusAlert to appear on success or failure of an asset being deleted.


src/data/constants/assets.js
```diff
@@ -0,0 +1,3 @@
+/* ASSETS RESPONSE MESSAGES */
+export const ERROR_FILE_DELETE = 'File could not be deleted.';
+export const SUCCESS_FILE_DELETE = 'File was successfully deleted.';
```
fysheets: Do we want the messages for responses to be more descriptive? Do we also want to store other static messages (i.e modal body text / confirmation button text / status alert text in this constants file as well, or do we prefer to have that inline? Is any of it reusable on other pages that we may do in the future?

src/data/reducers/assets.js
```diff
@@ -23,6 +23,13 @@ const list = (state = [], action) => {
 
 const status = (state = {}, action) => {
   switch (action.type) {
+    case assetActions.CLEAR_ASSETS_STATUS:
+      return {};
```
fysheets: I added this because the status message is what is being used to determine the styling of the StatusAlert -- there is a short amount of time where the alert will show up before the delete request has completed and returned a response, this could cause the styling to show incorrectly briefly (green first to red failure or vice versa).  By clearing the status after the close of the alert, the next time the alert is triggered it will show the warning/"Deleting asset" text before showing the correct status (yellow to red/green).

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -48,19 +53,40 @@ class AssetsTable extends React.Component {
...
+      }
+
+      focusAsset = this.state.displayAssetsList.find(asset => asset.row_key === (rowToDelete));
+      focusRowButton = this.trashcanRefs[focusAsset.id];
+    }
+
+    return focusRowButton;
+  }
+
+  addSupplementalTableElements() {
```
fysheets: Renamed this fuction as it now has more use cases than just adding the delete button -- will add lock button and now adds an ID for the row # that the data is on.  That row # is used for determining the next focusable element once the status alert is closed.

src/data/constants/assets.js
fysheets: Also this file was created because of the use of checking the status message to determine if the request being returned by the action creator was a success or failure, moving the constant string to a shared location reduces the number of places that will need to get changed in the future if we decide to change it. This checking of the message currently determines where focus should move from the StatusAlert.

efischer19: We're going to need to consider i18n with these strings as well - the "inline vs centralized" string storage decision will need to be made and standardized across the repo soon.

src/data/actions/assets.js
```diff
@@ -42,9 +46,12 @@ export const deleteAsset = (assetsParameters, assetId) =>
     clientApi.requestDeleteAsset(assetsParameters.courseId, assetId)
       .then((response) => {
         if (response.ok) {
-          dispatch(deleteAssetSuccess(assetId));
+          dispatch(deleteAssetSuccess(assetId, response, AssetsConstants.SUCCESS_FILE_DELETE));
```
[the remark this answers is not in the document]
fysheets: Should pose the question -- do we want to do this in this manner where the `SUCCESS_FILE_DELETE` action type is used in two different reducers, or do we want two dispatch separate `dispatch` calls and create a new `assetXHRSuccess` / `ASSET_XHR_SUCCESS` type? I feel like the way I did it isn't recommended even if it does technically work.  It might also be more clear if we do two separate dispatch calls? @arizzitano specific react/redux framework question

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -48,19 +53,40 @@ export class AssetsTable extends React.Component {
...
   }
 
-  addDeleteButton() {
-    const newAssetsList = this.props.assetsList.map((asset) => {
+  getNextFocusElementOnDelete(assetToDelete) {
+    let rowToDelete = assetToDelete.row_key;
+    let focusAsset = assetToDelete;
+    let focusRowButton = this.trashcanRefs[assetToDelete.id];
+
+    if (this.props.assetsStatus.text === AssetsConstants.SUCCESS_FILE_DELETE) {
```
dsjen: Is this function (in theory) ever called if this isn't the case? Also, the comparison of text is a bit worrisome especially in the case that the text is translated. Is there some sort of unique state/status/prop/code/you-get-the-gist that we could use instead?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -104,7 +138,7 @@ export class AssetsTable extends React.Component {
     );
   }
 
-  renderBody() {
+  renderModalBody() {
```
dsjen: I do think this is clearer. 🍨

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -115,11 +149,43 @@ export class AssetsTable extends React.Component {
...
   }
 
+  renderStatusAlert() {
+    const { assetsStatus } = this.props;
+    const deleteAssetName = this.state.assetToDelete.display_name;
+    let alertDialog = `Deleting ${deleteAssetName}`;
+    let alertType = 'info';
+
+    if (assetsStatus.text === AssetsConstants.ERROR_FILE_DELETE) {
+      alertDialog = `Unable to delete ${deleteAssetName}.`;
```
dsjen: Not something to worry about now, but we'll need for this to be internationalized too (e.g. string filled in via template) or some way to do this. CC: @efischer19

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -2,8 +2,11 @@ import React from 'react';
 import { mount } from 'enzyme';
 import { AssetsTable } from './index';
 
-let wrapper;
+import * as AssetsConstants from '../../data/constants/assets'; 
 
+const setAssetToDelete = (asset, wrapper) => {
```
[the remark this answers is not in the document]
fysheets: I figured we may want to test this in other tests moving forward so I moved it out to reduce the duplication that I was already doing - it could easily go back into each test, I wasn't sure which was preferable for testing.  I could see it making sense to have it in each since they're supposed to be individual tests, but this seemed cleaner for the duplication.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -27,9 +30,13 @@ const defaultProps = {
       sort: 'sort',
     },
   },
+  assetsStatus: {},
   deleteAsset: () => {},
+  clearAssetsStatus: () => {},
 };
 
+let wrapper;
```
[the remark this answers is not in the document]
fysheets: In all fairness this was predefined when I got into this file - I just had to move it's location below the `setAssetToDelete` from above.  I don't mind changing it from the global to individual declarations, I just wasn't the one to define it this way initially so I'm not sure if it was done for a reason?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -41,3 +48,97 @@ describe('<AssetsTable />', () => {
     expect(wrapper.find('tr')).toHaveLength(defaultProps.assetsList.length + 1);
   });
 });
+
+describe('it displays alert properly', () => {
+  it('renders info alert with blank status', () => {
```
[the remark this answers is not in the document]
fysheets: Good catch! It does not check the status message, it is misnamed.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -3,38 +3,43 @@ import PropTypes from 'prop-types';
...
 import Button from '@edx/paragon/src/Button';
 import Modal from '@edx/paragon/src/Modal';
+import StatusAlert from '@edx/paragon/src/StatusAlert';
 import classNames from 'classnames';
 import { connect } from 'react-redux';
 
 import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
-import { deleteAsset } from '../../data/actions/assets';
-
+import { deleteAsset, clearAssetsStatus } from '../../data/actions/assets';
```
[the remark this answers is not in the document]
fysheets: I was not aware of the preference, but I'm all for keeping some sort of convention across repos, I'll swap that to be alphabetical.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -48,19 +53,40 @@ export class AssetsTable extends React.Component {
     });
   }
 
-  addDeleteButton() {
-    const newAssetsList = this.props.assetsList.map((asset) => {
+  getNextFocusElementOnDelete(assetToDelete) {
```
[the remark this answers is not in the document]
fysheets: Done! And agreed, it is clearer.

src/AssetsPage/AssetsTable/index.jsx
fysheets: Yes, it will still be called because regardless of the response status we need to know where to send focus when the status alert is closed out of. The text comparison is using the response body text, not text that will ever been shown to the user / internationalized. (The ones that now live in the constants/assets.js file). But I understand the concern, I'll test out if I can use response code or something else here.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -48,19 +53,40 @@ export class AssetsTable extends React.Component {
...
+    if (this.props.assetsStatus.text === AssetsConstants.SUCCESS_FILE_DELETE) {
+      if (rowToDelete > 0) {
+        rowToDelete -= 1;
+      }
+
+      focusAsset = this.state.displayAssetsList.find(asset => asset.row_key === (rowToDelete));
+      focusRowButton = this.trashcanRefs[focusAsset.id];
+    }
+
+    return focusRowButton;
```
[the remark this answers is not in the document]
fysheets: Definitely -- good catch on some clean up.

src/AssetsPage/AssetsTable/index.jsx
fysheets: 👍 Agreed. We'll also need it for the Modal body text & buttons.

src/AssetsPage/AssetsTable/index.jsx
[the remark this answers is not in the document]
fysheets: Actually, after reading more of your comments -- it's determining where to focus after closing the StatusAlert that appears on delete. I'm not sure if the `OnDeleteClose` is clear about that, but if we're going to be using the status alert for multiple alert types (delete/copy URL/...) I'm not sure if `OnAlertClose` will be clear either because that will have different rules, though I guess we could maybe extend that function to encompass all the rules related to the alert and focus?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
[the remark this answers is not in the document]
dsjen: Ah, I see! I guess I'd personally not do it globally and would probably make that change.

src/AssetsPage/AssetsTable/index.jsx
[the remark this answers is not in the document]
dsjen: Ok! Either way works for me, but I just wanted to check on that just for my own sake mostly. :)

src/AssetsPage/AssetsTable/index.jsx
dsjen: Oh, I see! I think that `assetsStatus` has a `text` field is what confused me. That's your status code, right? I don't want to bloat your PR, but I think that if `text` were renamed to `statusCode` or `type` (or something else), it would be clearer. To me, `text` indicates that it's displayed to the user.

src/AssetsPage/AssetsTable/index.jsx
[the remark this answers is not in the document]
dsjen: Ah, I see! Maybe the current name makes the most sense actually. I'm reading it again and finding it to be clear!

src/data/reducers/assets.js
```diff
@@ -23,6 +23,13 @@ const list = (state = [], action) => {
...
+    case assetActions.DELETE_ASSET_SUCCESS:
+      return {
+        response: action.response,
+        text: action.text,
```
dsjen: Recap of our in-person conversation:
* we'll use the existing `type` instead of text to compare against
* the `type` constants will be moved into a separate file

Thanks for the discussion!

src/data/constants/actionTypes.js
```diff
@@ -0,0 +1,9 @@
+/* ASSETS ACTION TYPES */
```
dsjen: 😍

dsjen:
🙌  🍰  💯

PR #41, opened 2017-10-19
(WIP) mroytman/sortable assets headings

MichaelRoytman:


src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -38,6 +79,41 @@ class AssetsTable extends React.Component {
...
+    const column = this.state.tableColumns.find(col => (col.key === columnKey));
+    let newDirection;
+
+    if (this.state.sortState.columnKey === columnKey) {
+      newDirection = this.state.sortState.direction === 'descending' ? 'ascending' : 'descending';
+      column.sortDirection = newDirection;
+
+      // then we want to make the appropriate API call
+    } else {
+      // is this the best way to do a line break?
```
MichaelRoytman: Re: how to do line breaks the best way in React

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -69,20 +183,19 @@ class AssetsTable extends React.Component {
...
-    // TODO shift focus to banner, something like:
-    // this.setState({
-    //  elementToFocusOnModalClose: <alertStatus>
-    // })
-    // this.state.elementToFocusOnModalClose.focus();
+  renderTableColumnButton(column) {
+    return (column.sortable ?
+      <Button
+        // LINE BREAK?
+        display={<span> <span> {column.label} </span> {this.getSortIcon(column.sortDirection)} </span>}
```
MichaelRoytman: Re: how to do line breaks the best way in React

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -38,6 +79,41 @@ class AssetsTable extends React.Component {
...
+        (col.key === this.state.sortState.columnKey));
+
+      // can we have a better way to do this that's less ambiguous than relying on the default?
+      oldColumn.sortDirection = '';
+      newDirection = 'ascending';
+      column.sortDirection = newDirection;
+
+      // then we want to make the appropriate API call
+      // who should handle this? AssetsTable? What happens if a user has a filter selected and a sort? Does this integrate
+      // with what's happening in AssetsPage? I need to look into this further.
```
MichaelRoytman: I need to consider this!

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -38,6 +79,41 @@ class AssetsTable extends React.Component {
...
+      column.sortDirection = newDirection;
+
+      // then we want to make the appropriate API call
+    } else {
+      // is this the best way to do a line break?
+      const oldColumn = this.state.tableColumns.find(col =>
+        (col.key === this.state.sortState.columnKey));
+
+      // can we have a better way to do this that's less ambiguous than relying on the default?
+      oldColumn.sortDirection = '';
```
MichaelRoytman: This is referring to the switch statement in getSortIcon; I'd rather there be a case for 'other' or 'none' or something, but it would be identical to the default case, I think. Unless the default case is ''?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -18,16 +18,57 @@ class AssetsTable extends React.Component {
...
+        },
+        {
+          // label: this.renderTableColumnButton('Delete Asset', false, ''),
+          label: 'Delete Asset',
+          key: 'delete_asset',
+          sortable: false,
+        },
+      ],
+      sortState: {
+        // should this be dynamic instead of hardcoded? first sortable element of the columns? how to do that?
```
MichaelRoytman: This is assuming that the default initial state is to sort the first sortable element of the column in descending order.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -18,12 +17,52 @@ class AssetsTable extends React.Component {
       modalOpen: false,
       assetToDelete: {},
       elementToFocusOnModalClose: {},
+      tableColumns: [
+        // keyed object? or just search through the object when I need it? performance?
```
efischer19: Performance-wise, this is O(n) for linear search vs O(1) for hash lookup in a keyed object right?

I doubt it matters much when n is 4, but if it's not difficult to do I think the keyed object is the "right" decision here.

src/AssetsPage/AssetsTable/index.jsx
efischer19: I could see date being the default sorting as well, but that's probably a question to be determined by product later on in this process.

src/AssetsPage/AssetsTable/index.jsx
efischer19: Do you mean in code, or inside of display elements? In code, as long as we're somewhat consistent and readable I don't have a strong opinion. In display elements, I think `span` elements are correct, according to a quick google search.

src/AssetsPage/AssetsTable/index.jsx
efischer19: I'm okay with explicitly specifying '' as the default case - the 3 icons choices (`fa-sort`, `fa-sort-asc`, and `fa-sort-desc`) seem to correspond with those options well.

src/AssetsPage/AssetsTable/index.jsx
efischer19: Good question! Having AssetsTable in charge seems like a good initial solution, but we should probably discuss as a team.

src/AssetsPage/AssetsTable/index.jsx
fysheets: I think 'none' or something that isn't empty string is more clear when reading the code (say if there's someone who hasn't seen this discussion) and it could either be left to fall into the default or be part of the switch right before the default so it still falls through to the default anyway.  It's a bit more to write/be cognizant of when writing for the future however to keep things unified.  I tend to lean towards readability, but am open to be swayed :)

package.json
```diff
@@ -61,6 +61,9 @@
       "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.js",
       "\\.(css|scss)$": "identity-obj-proxy"
     },
+    "collectCoverageFrom": [
```
MichaelRoytman: This fixes the coverage issue where only files with corresponding test files were being used for coverage calculations.

src/AssetsPage/AssetsFilters/index.jsx
```diff
@@ -21,7 +29,7 @@ const ASSET_TYPES = [
   },
 ];
 
-const AssetsFilters = ({ assetsParameters, updateFilter }) => (
+export const AssetsFilters = ({ assetsParameters, updateFilter }) => (
```
MichaelRoytman: This might be premature, but it'll be necessary when testing the AssetsFilter, since we want to test the unwrapped component. It actually was meant for a separate testing PR but ended up here. I'm happy to remove it if it doesn't make sense here.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -5,11 +5,15 @@ import { AssetsTable } from './index';
 import { assetActions } from '../../data/constants/actionTypes';
 
 const setAssetToDelete = (asset, wrapper) => {
-  wrapper.setState({ assetToDelete: asset });
+  wrapper.setState({ deletedAsset: asset });
```
MichaelRoytman: This just modifies the function to incorporate the change we made with assetToDelete v. deletedAsset in the AssetTable.

src/AssetsPage/index.jsx
```diff
@@ -20,8 +20,7 @@ class AssetsPage extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    if (prevProps.assetsParameters.assetTypes !== this.props.assetsParameters.assetTypes) {
-      // if filters changed, update the assetsList
+    if (prevProps.assetsParameters !== this.props.assetsParameters) {
```
MichaelRoytman: I've changed this to be a condition on the assetsParameters changing at all, since we want to include changes not just to the assetTypes but also to the sort state.

src/AssetsPage/index.jsx
```diff
@@ -37,17 +36,14 @@ class AssetsPage extends React.Component {
...
-        />
+        <WrappedAssetsFilters />
+        <WrappedAssetsTable />
       </div>
     );
   }
 }
 
 AssetsPage.propTypes = {
-  assetsList: PropTypes.arrayOf(PropTypes.object).isRequired,
```
MichaelRoytman: The AssetsTable was subscribed to the assetsList slice of the Redux state and was receiving it as a prop from the AssetsPage, so I've removed assetsList from here in favor of the AssetsTable subscribing to the assetsList itself.

src/AssetsPage/index.jsx
```diff
@@ -56,7 +52,6 @@ AssetsPage.propTypes = {
 
 const WrappedAssetsPage = connect(
   state => ({
-    assetsList: state.assets.list,
```
MichaelRoytman: The AssetsTable was subscribed to the assetsList slice of the Redux state and was receiving it as a prop from the AssetsPage, so I've removed assetsList from here in favor of the AssetsTable subscribing to the assetsList itself.

src/AssetsPage/index.jsx
MichaelRoytman: See below.

src/api/client.test.jsx
```diff
@@ -18,12 +18,13 @@ describe('API client requestAssets', () => {
     requestAssets(COURSE_ID, {
       page: 2,
       pageSize: 10,
-      sort: 'asc',
+      sort: 'content_type',
```
MichaelRoytman: Just fixing this up, since the API expects sort to be the name of the attribute on which to sort on and direction to be the direction ('asc' or 'desc').

src/data/constants/actionTypes.js
```diff
@@ -6,4 +6,5 @@ export const assetActions = {
   DELETE_ASSET_SUCCESS: 'DELETE_ASSET_SUCCESS',
   ASSET_XHR_FAILURE: 'ASSET_XHR_FAILURE',
   CLEAR_ASSETS_STATUS: 'CLEAR_ASSETS_STATUS',
+  SORT_UPDATE: 'SORT_UPDATE',
```
MichaelRoytman: Self explanatory, but we now have a new action type that signifies a change in the selected sort.

src/data/actions/assets.js
```diff
@@ -11,6 +11,8 @@ export const getAssets = assetsParameters =>
     clientApi.requestAssets(assetsParameters.courseId, {
       page: assetsParameters.page,
       assetTypes: assetsParameters.assetTypes,
+      sort: assetsParameters.sort,
+      direction: assetsParameters.direction,
```
MichaelRoytman: I'm now passing in the sort and the direction to the API.

src/data/actions/assets.js
```diff
@@ -20,6 +22,12 @@ export const filterUpdate = (filterKey, filterValue) => ({
   data: { [filterKey]: filterValue },
 });
 
+export const sortUpdate = (sort, direction) => ({
```
MichaelRoytman: sortUpdate is an action creator that creates an action of type SORT UPDATE and with data that is an object containing the sort and direction.

src/data/reducers/assets.js
```diff
@@ -7,7 +7,8 @@ const initialState = {
   page: 0,
   pageSize: 50,
   assetTypes: {},
-  sort: 'sort',
+  sort: 'date_added',
+  direction: 'desc',
```
MichaelRoytman: I figure the initial state of the assetParameters would be a good place to stuff in the default sort state: date_added descending.

src/data/reducers/assets.js
```diff
@@ -40,20 +41,17 @@ const status = (state = {}, action) => {
   }
 };
 
-const types = (state = {}, action) => {
+const parameters = (state = initialState, action) => {
```
MichaelRoytman: I have absorbed the old types reducer into a new parameters reducer, which handles changes to the request parameters.

src/data/reducers/assets.js
```diff
@@ -40,20 +41,17 @@ const status = (state = {}, action) => {
...
   switch (action.type) {
     case assetActions.FILTER_UPDATED:
+      return { ...state, assetTypes: { ...state.assetTypes, ...action.data } };
```
MichaelRoytman: assetTypes is an object of key-value pairs where the key is filter name (e.g. OTHER, Images, etc.), and the value is whether or not the filter is selected (true or false, respectively). Since the assetTypes object should store all selected filters, we have to expand the current assetTypes and tack on the data of the action.

src/api/client.js
```diff
@@ -3,6 +3,13 @@ import Cookies from 'js-cookie';
 
 import endpoints from './endpoints';
 
+// terrible name, but it'll do for the moment
```
MichaelRoytman: The comment stands. I can't think of an elegant name to describe what this dictionary is for. Perhaps it's because the problem it attempts to solve is an ugly one.

Basically, the underlying assets database has the following names for asset attributes we care about: displayname, contentType, and uploadDate. Notice that displayname is not cased, contentType is camel-cased, and uploadDate is camel-cased but doesn't match the name returned by the API (date_added).

The API handles the conversion of date_added to uploadDate and display_name to displayname but not content_type to contentType.
The code for that is [here](https://github.com/edx/edx-platform/blob/master/cms/djangoapps/contentstore/views/assets.py#L279).

The actual API response you get [back](https://github.com/edx/edx-platform/blob/master/cms/djangoapps/contentstore/views/assets.py#L559) when requesting assets uses display_name, content_type, and date_added, so you can't just use the asset attribute names returned by the API to request a sort type, since the contentType case isn't handled. 

We could, alternatively, add a condition to the assets code to handle contentType, but I kind of just wanted to deal with it here so that it's completely explicit. I had to spend a good amount of time figuring out why this wasn't working, so I would like to avoid someone having to do the same later.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
   return Object.keys(assetTypes).filter(key => assetTypes[key]);
 }
 
-export function requestAssets(courseId, { page = 0, pageSize = 50, sort = 'sort', assetTypes = {} }) {
```
MichaelRoytman: The API has these same defaults[ built in](https://github.com/edx/edx-platform/blob/master/cms/djangoapps/contentstore/views/assets.py#L32), so I didn't want them specified here as well.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
...
-  const assetTypesToFilter = assetTypesFromState(assetTypes);
+function getDatabaseAttributesFromAssetAttributes(sort) {
+  return assetAttributesToDatabaseAttributes[sort];
+}
+
+function objectIsNonEmpty(object) {
+  return Object.keys(object).length > 0;
+}
+
+export function requestAssets(courseId, params) {
```
MichaelRoytman: The function now takes an object called params that contains the request parameters.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
...
+  const assetTypesToFilter = assetTypesFromState(params.assetTypes);
```
MichaelRoytman: This is the same as before.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
...
+  const sortType = getDatabaseAttributesFromAssetAttributes(params.sort);
```
MichaelRoytman: This converts the sortType from the attribute names returned by the API to the attribute names expected by the API.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
...
+
+  const parameters = {
```
MichaelRoytman: This handles the camel case v. snake case problem between our Redux initial state and the names expected by the API.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
...
+    ...params,
+    sort: sortType,
+    asset_type: (objectIsNonEmpty(assetTypesToFilter) ? assetTypesToFilter : undefined),
+    assetTypes: undefined,
+    page_size: params.pageSize,
+    pageSize: undefined,
+  };
+
+  const requestString = Object.keys(parameters).reduce((memo, key) => { if (parameters[key]) { memo.push(`${key}=${parameters[key]}`); } return memo; }, []).join('&');
```
MichaelRoytman: Because we no longer specify defaults, we have to dynamically generate the request string depending upon the contents of the parameters object.

src/api/client.js
```diff
@@ -18,10 +25,31 @@ function assetTypesFromState(assetTypes) {
...
+}
+
+export function requestAssets(courseId, params) {
+  const assetTypesToFilter = assetTypesFromState(params.assetTypes);
+  const sortType = getDatabaseAttributesFromAssetAttributes(params.sort);
+
+  const parameters = {
+    ...params,
+    sort: sortType,
+    asset_type: (objectIsNonEmpty(assetTypesToFilter) ? assetTypesToFilter : undefined),
```
MichaelRoytman: Sigh. The value of assetTypes in Redux's parameters state is {}. The reduce function at line 49 ends up tacking on the substring "assetTypes=", since parameters[assetType] is true, given that assetType = {}. Therefore, I had to basically add the assertion that we only add assetTypes to the request string if parameters[key].legth > 0 (i.e. assetTypes exists but isn't just {}). Should we just change the default initialState in the Redux store to undefined?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,109 @@ import { connect } from 'react-redux';
...
 
 export class AssetsTable extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      displayAssetsList: [],
       modalOpen: false,
       statusAlertOpen: false,
       assetToDelete: {},
+      deletedAsset: {},
```
MichaelRoytman: deletedAsset stores a copy of assetToDelete after the asset has been deleted from the Redux store. We need this information to populate the status alert with data about the already deleted asset but didn't want to rely on a reference to a now-deleted asset. It also allows us to clear the value of assetToDelete after the delete, keeping the semantics of their names correct.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,109 @@ import { connect } from 'react-redux';
...
+      deletedAssetIndex: null,
```
MichaelRoytman: The index of the asset that was deleted in the original assetList. This is used to determine where focus should go after the status alert is closed.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,109 @@ import { connect } from 'react-redux';
...
       elementToFocusOnModalClose: {},
-      statusAlertRef: {},
+      columnSortState: {
+        display_name: 'none',
+        content_type: 'none',
+        date_added: 'desc',
+      },
     };
 
+    this.columns = {
```
MichaelRoytman: I've moved the columns from the render up into an instance variable and added the columnSortable key to specify which columns are sortable.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,109 @@ import { connect } from 'react-redux';
...
     this.state = {
-      displayAssetsList: [],
       modalOpen: false,
       statusAlertOpen: false,
       assetToDelete: {},
+      deletedAsset: {},
+      deletedAssetIndex: null,
       elementToFocusOnModalClose: {},
-      statusAlertRef: {},
+      columnSortState: {
```
MichaelRoytman: This keeps track of which column is sorted.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
       modalOpen: false,
       statusAlertOpen: false,
       assetToDelete: {},
+      deletedAsset: {},
+      deletedAssetIndex: null,
       elementToFocusOnModalClose: {},
-      statusAlertRef: {},
     };
 
+    this.columns = {
```
MichaelRoytman: I've moved the columns up into an instance variable instead of in the render so that I could more easily map over them an add an onSortClick in the render function.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
     this.deleteAsset = this.deleteAsset.bind(this);
     this.addSupplementalTableElements = this.addSupplementalTableElements.bind(this);
     this.closeModal = this.closeModal.bind(this);
     this.closeStatusAlert = this.closeStatusAlert.bind(this);
     this.renderStatusAlert = this.renderStatusAlert.bind(this);
-
-    this.trashcanRefs = {};
   }
 
-  componentDidMount() {
```
MichaelRoytman: I removed componentDidMount in favor of computing the additional table elements on the fly in the render method, since we wanted to not keep two copies of essentially the same data in memory between the Redux store and the component state (in this.props.assetsList and this.state.displayAssetsList, respectively).

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
-  }
+  // componentDidUpdate -> runs after render;
+  // use it for focus
+  // check if set to null
+
+  onSortClick(columnKey) {
+    const sortedColumn = this.props.assetsParameters.sort;
+    const sortedDirection = this.props.assetsParameters.direction;
 
-  componentDidUpdate(prevProps) {
```
MichaelRoytman: Since the component now subscribes to the assetsList slice of the Redux store, it will re-render whenever the Redux state changes anyway, so I figure this method is no longer necessary.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
-    if (prevProps.assetsList.length !== this.props.assetsList.length) {
-      this.addSupplementalTableElements();
+    let newDirection = 'desc';
+
+    if (sortedColumn === columnKey) {
+      newDirection = sortedDirection === 'desc' ? 'asc' : 'desc';
     }
+
+    this.props.updateSort(columnKey, newDirection);
```
MichaelRoytman: Call updateSort function, which dispatches an updateSort action creator.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
   }
 
-  onDeleteClick(assetId) {
-    const assetToDelete = this.state.displayAssetsList.find(asset => (asset.id === assetId));
+  onDeleteClick(index) {
```
MichaelRoytman: onDeleteClick now gets the index of the asset in this.props.assetsList.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
+    let focusAsset = this.state.deletedAsset;
 
     switch (assetsStatus.type) {
       case assetActions.DELETE_ASSET_SUCCESS:
-        if (rowToDelete > 0) {
-          rowToDelete -= 1;
+        if (deletedIndex > 0) {
+          deletedIndex -= 1;
         }
-        focusAsset = this.state.displayAssetsList.find(asset => asset.row_key === (rowToDelete));
```
MichaelRoytman: RIP displayAssetsList. Long live this.props.assetsList!

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -91,33 +125,44 @@ export class AssetsTable extends React.Component {
...
-
-    this.setState({
-      displayAssetsList: newAssetsList,
-    });
+    return newAssetsList;
   }
 
   closeModal() {
-    this.setState({ modalOpen: false });
     this.state.elementToFocusOnModalClose.focus();
```
MichaelRoytman: Focus, then set all the state so that we only set state once here.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -91,33 +125,44 @@ export class AssetsTable extends React.Component {
...
+      deletedAsset: {},
+      deletedAssetIndex: null,
       statusAlertOpen: false,
     });
-    this.getNextFocusElementOnDelete(this.state.assetToDelete).focus();
-    this.props.clearAssetsStatus();
   }
 
   deleteAsset() {
+    const deletedAsset = { ...this.state.assetToDelete };
```
MichaelRoytman: Set deletedAsset to be a copy of this.state.assetToDelete.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -194,25 +239,14 @@ export class AssetsTable extends React.Component {
...
-              key: 'delete_asset',
-            },
-          ]}
-          data={this.state.displayAssetsList}
+          columns={Object.keys(this.columns).map(columnKey => ({
+            ...this.columns[columnKey],
+            onSort: () => this.onSortClick(columnKey),
+          }))}
+          data={this.addSupplementalTableElements(this.props.assetsList)}
+          tableSortable
```
MichaelRoytman: These are new props of the sortable Table in Paragon.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,354 @@ const defaultProps = {
...
+
+        wrapper = mount(
+          <AssetsTable
+            {...defaultProps}
+            updateSort={updateSortSpy}
+            assetsParameters={sortableAssetParameters}
+          />,
+        );
+
+        const sortButtons = wrapper.find('button').filterWhere(button => (
```
MichaelRoytman: This finds all sortable column buttons.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,354 @@ const defaultProps = {
...
+        sort: 'date_added',
+        direction: 'asc',
+      },
+      {
+        sort: 'date_added',
+        direction: 'desc',
+      },
+    ];
+
+    const assertValues = [
```
MichaelRoytman: Basically, these are values needed by each test.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,360 @@ const defaultProps = {
...
+      },
+      {
+        buttonIndexToClick: 0,
+        buttonToClickText: 'Name',
+        onSortColumnParameter: 'display_name',
+        onSortDirectionParameter: 'desc',
+      },
+    ];
+
+    sortableAssetProps.forEach((sortableAssetProp, index) => {
```
MichaelRoytman: For each sortableAssetProp, we dynamically generate a test that starts with a sort state that's specified by sortableAssetProp.sort and sortableAssetProp.direction.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
+    it('renders', () => {
+      expect(modal).toHaveLength(1);
+    });
+    it('is closed by default', () => {
+      expect(modal.hasClass('modal-open')).toEqual(false);
+      expect(wrapper.state('modalOpen')).toEqual(false);
+    });
+    it('opens when trash button clicked; modalOpen state is true', () => {
+      const trashButtons = wrapper.find('button').filterWhere(button => button.hasClass('fa-trash'));
+      expect(trashButtons).toHaveLength(defaultProps.assetsList.length);
```
MichaelRoytman: I should remove this condition, since it's already tested (it's the second test of the render suite).

src/api/client.js
fysheets: Will it break anything if we do set it to `undefined` in the redux store instead of `{}`? If not, I'm all for uniformity.  If yes, I think this is fine for now.

src/api/client.js
fysheets: For my own curiosity - is it a convention to call it `memo` when performing this sort of task?

src/data/reducers/assets.js
fysheets: I'm glad this now follows the same structure of the other reducers, it makes it more clear.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
fysheets: Nit -- can we update the const name to `setDeletedAsset`?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
-    expect(wrapper.find('tr')).toHaveLength(defaultProps.assetsList.length + 1);
+  describe('renders', () => {
+    beforeEach(() => {
+      wrapper = mount(
+        <AssetsTable
+          {...defaultProps}
+        />,
+      );
+    });
+    it(' correct number of assets', () => {
```
fysheets: Nit - Does the space need to be here before the ` correct`?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
+  });
+  describe('onSortClick', () => {
+    let updateSortSpy;
+
+    beforeEach(() => {
+      updateSortSpy = jest.fn();
+    });
+
+    const sortableAssetProps = [
+      {
```
fysheets: Why is sort the same for each of these objects?

Edit -- I see the reasoning in the comments below.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
+        sortButtons.at(testValues.buttonIndexToClick).simulate('click');
+
+        expect(updateSortSpy).toHaveBeenCalledTimes(1);
+        expect(updateSortSpy).toHaveBeenLastCalledWith(
+          testValues.onSortColumnParameter,
+          testValues.onSortDirectionParameter,
+        );
+      });
+    });
+  });
```
fysheets: Nice test - I was able to understand it clearly when looking at the updated file as a whole.  It also covers a lot of different testing features that we should leverage when creating future tests.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
+        <AssetsTable
+          {...defaultProps}
+          deleteAsset={deleteAssetSpy}
+        />,
+      );
+
+      const deleteButtons = wrapper.find('button');
+      const trashButtons = deleteButtons.filterWhere(button => button.hasClass('fa-trash'));
+      const deleteButton = deleteButtons.filterWhere(button =>
+        button.matchesElement(<button>Yes, delete.</button>));
```
fysheets: Curiosity - Can this be done with finding the buttons within the modal explicitly and then finding the one with `btn-primary` style?

fysheets: Actually below you do search from the modal buttons specifically: https://github.com/edx/studio-frontend/pull/41/files#diff-ca1fa9e3aa238807437edc787d4e8f9dR281 - still using the specific HTML tags.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
+      modal = wrapper.find('[role="dialog"]');
+      closeButton = modal.find('button').filterWhere(button => button.matchesElement(<button><span>&times;</span></button>));
+    });
+
+    it('moves from modal to trashcan on modal close', () => {
+      trashButtons.at(0).simulate('click');
+
+      expect(closeButton.at(0).matchesElement(document.activeElement)).toEqual(true);
+
+      closeButton.at(0).simulate('click');
```
fysheets: Is the `at(0)` necessary if you're filtering on the buttons within the modal specifically? There should only be one close button from that group of buttons.  Or is the `at(0)` supposed to be an extra validation?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -24,31 +28,335 @@ const defaultProps = {
...
+      deleteButton.at(0).simulate('click');
+
+      expect(closeStatusAlertButton.at(0).matchesElement(document.activeElement)).toEqual(true);
+    });
+    it('moves to correct asset trashcan icon after first asset deleted', () => {
+      const assetToDeleteId = defaultProps.assetsList[0].id;
+
+      mockDeleteAsset = jest.fn(() => {
+        wrapper.setProps({ assetsList: defaultProps.assetsList.filter(asset => asset.id !== assetToDeleteId), assetsStatus: { type: 'DELETE_ASSET_SUCCESS' } });
+      });
```
fysheets: Can this `mockDeleteAsset` move to a more general location? It's duplicated below as well, or is there a reason we would want to keep it separated out?

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "license": "AGPL-3.0",
   "dependencies": {
-    "@edx/paragon": "^0.2.0",
+    "@edx/paragon": "git://github.com/edx/paragon.git#ae1e93d873c5f93f56446f5f35f1624bc81ebca5",
```
fysheets: Note* this is waiting on a Paragon merge to update to the latest version number prior to merging.

src/api/client.js
MichaelRoytman: I don't think so? I think it's just the name Ari gave it, since she helped me with this part. I think I'll rename it to request, just so it's a little more clear.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
MichaelRoytman: Do you think there's a way it could be clearer? I see that the sortableAssetProps might have been a little confusing.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
MichaelRoytman: What's the advantage of that?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
MichaelRoytman: I checked, and it doesn't appear to be necessary. I think I got in the habit of doing that for instances where the ReactWrapper had more than one node in it, in which case it would complain with something like ```Method “simulate” is only meant to be run on a single node. 4 found instead.```. I think it might actually be better to leave it out, since it allows for the test to fail if more than the expected number of nodes are in the wrapper. If we find more than one closeButton, that's probably a problem, and doing closeButton.at(0) would mask that.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
fysheets: This is more of a nit for the styling search -- the finding of buttons on the modal specifically than out of all the buttons in the off chance another "Yes, delete." button gets added anywhere else in the DOM. This is fine for now, but contextually it made sense to me to search for the appropriate button from within the modal here since that's where it lives (and you had done it that way in the other test, so I guess some uniformity too haha).

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
MichaelRoytman: Ideally, I'd move it up into the beforeEach, but because assetToDeleteId differs between the tests, I need to re-initialize the function so that it includes the value of assetToDeleteId that's specific to that test. Can you think of a way to abstract that away?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -9,61 +9,94 @@ import { connect } from 'react-redux';
...
     this.closeStatusAlert = this.closeStatusAlert.bind(this);
     this.renderStatusAlert = this.renderStatusAlert.bind(this);
-
-    this.trashcanRefs = {};
   }
 
-  componentDidMount() {
-    this.addSupplementalTableElements();
-  }
+  // componentDidUpdate -> runs after render;
```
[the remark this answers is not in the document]
MichaelRoytman: I guess I wanted it to serve more as a question. Should we deal with this in this PR? I think the focus is working correctly without it.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
fysheets: I think it's fine, seeing it as a whole makes sense.  Wouldn't worry about it too much at this point.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
MichaelRoytman: Meaning something like this?

`wrapper.find('[role="dialog"]').filterWhere(button => button.matchesElement(<button>Yes, delete.</button>));`

Or do you prefer?

`wrapper.find('[role="dialog"]').filterWhere(button => button.hasClass('btn-primary'));`?

I feel like I prefer the first one since it selects for a delete specific button in the modal, as opposed to a primary button in the modal, which may not be the delete? I could go either way, though.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
efischer19: I'm a little concerned about having 2 dictionaries that are implicitly linked. Could we iterate over a single, more complex object to avoid using `index`? Something like:

```
const testData = [
  {
    sortProps: {
      sort: 'date_added',
      direction: 'desc'
    },
    expectedValues: {
      <appropriate key: value pairs>
    }
  },
  ...
]
```

Then the test loop becomes `testData.forEach`. Just an idea, if this works currently and it's difficult to switch don't sweat it.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
efischer19: Could we curry it? Define a function (in some utility location) that takes an id, and returns this no-parameters function?

Just spitballing here, I'm okay with it as-is.

src/AssetsPage/AssetsTable/index.jsx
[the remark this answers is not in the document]
efischer19: If focus is working correctly I say :shipit:

src/data/actions/assets.js
```diff
@@ -45,3 +53,4 @@ export const deleteAsset = (assetsParameters, assetId) =>
 export const clearAssetsStatus = () =>
   dispatch =>
     dispatch({ type: assetActions.CLEAR_ASSETS_STATUS });
+
```
efischer19: nit: random newline

MichaelRoytman:
@efischer19 Thanks for the suggestions for currying the function and creating a larger testData object. Those were helpful, and I think they made the test file more clean.

I think this should be good to go, if everyone is satisfied.

PR #42, opened 2017-10-19
fix AssetsPage to import WrappedAssetsTable correctly

MichaelRoytman:


MichaelRoytman:
For reference, the issue was [this line](https://github.com/edx/studio-frontend/blob/e0761a7462c07e08210645743b8448320646cff6/src/AssetsPage/index.jsx#L4) in the testing infrastructure set-up PR. I had made the unconnected/unwrapped AssetsTable a named export of the AssetsTable module so that I could test the AssetsTable independent of the Redux store. I, therefore, changed (incorrectly) the import in AssetsPage to { AssetsTable } without realizing that, I did, in this case, actually want the wrapped AssetsTable. I changed it to be WrappedAssetsTable. What was happening was that React was complaining that it didn't have a deleteAsset function, because it receives that function as a property from the Redux store.

PR #44, opened 2017-10-24
change Button display prop to label

MichaelRoytman:


MichaelRoytman:
This follows from https://github.com/edx/paragon/pull/54.

dsjen:
Does paragon need to be bumped too?

PR #46, opened 2017-10-26
Incorporate Studio Data into Redux Store

fysheets:
[EDUCATOR-1529](https://openedx.atlassian.net/browse/EDUCATOR-1529)

Goal is to incorporate studio data into SFE, current idea is to include the course context onto the page with a script tag and leverage that within SFE's redux store throughout the app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,9 @@
     ],
     "transformIgnorePatterns": [
       "/node_modules/(?!(@edx/paragon)/).*/"
-    ]
+    ],
+    "globals": {
+      "courseContext": {}
+    }
   }
 }
--- a/public/index.html
+++ b/public/index.html
@@ -2,6 +2,21 @@
 <html>
   <head></head>
   <body>
+    <script type="text/javascript" id='courseContext'>
+/* eslint-disable no-unused-vars */
+/* eslint-disable no-var */
+/* eslint-disable quotes */
+var courseContext = {
+  lang: "fr",
+  url_name: "course",
+  name: "edX Demonstration Course",
+  display_course_number: "",
+  num: "DemoX",
+  org: "edX",
+  id: "course-v1:edX+DemoX+Demo_Course",
+  revision: "",
+};
+    </script>
     <div id="root"></div>
   </body>
 </html>
--- a/public/index.html
+++ b/public/index.html
@@ -2,6 +2,10 @@
 <html>
   <head></head>
   <body>
+    <!-- This script tag is being included for testing within -->
+    <!-- the studio frontend container locally.  This value will be -->
+    <!-- dynamically passed in from Studio backend when SFE is -->
+    <!-- run from within Platform/Studio. -->
     <script type="text/javascript" id='courseContext'>
 /* eslint-disable no-unused-vars */
 /* eslint-disable no-var */
@@ -15,6 +19,7 @@ var courseContext = {
   org: "edX",
   id: "course-v1:edX+DemoX+Demo_Course",
   revision: "",
+  base_url: "http://localhost:18010",
 };
     </script>
     <div id="root"></div>
--- a/src/AssetsPage/AssetsTable/index.jsx
+++ b/src/AssetsPage/AssetsTable/index.jsx
@@ -146,7 +146,9 @@ export class AssetsTable extends React.Component {
   deleteAsset() {
     const deletedAsset = { ...this.state.assetToDelete };
 
-    this.props.deleteAsset(this.props.assetsParameters, this.state.assetToDelete.id);
+    this.props.deleteAsset(this.props.assetsParameters,
+      this.state.assetToDelete.id,
+      this.props.courseDetails);
 
     this.setState({
       assetToDelete: {},
@@ -226,24 +228,28 @@ export class AssetsTable extends React.Component {
   }
 
   render() {
-    return (!this.props.assetsList.length) ? (
-      <span>Loading....</span>
-    ) : (
-      <div>
-        {this.renderStatusAlert()}
-        <Table
-          columns={Object.keys(this.columns).map(columnKey => ({
-            ...this.columns[columnKey],
-            onSort: () => this.onSortClick(columnKey),
-          }))}
-          data={this.addSupplementalTableElements(this.props.assetsList)}
-          tableSortable
-          defaultSortedColumn="date_added"
-          defaultSortDirection="desc"
-        />
-        {this.renderModal()}
-      </div>
-    );
+    // TODO: Add UI for when there is nothing in the list and we have a status returned from the API
+    // TODO: http://fhtwd0.axshare.com/#g=1&p=files-and-uploads-empty
+    // Only show Loading when the list is empty AND the status from the APIs are empty
+    return (this.props.assetsList.length === 0 &&
+      Object.keys(this.props.assetsStatus).length === 0) ? (
+        <span>Loading....</span>
+      ) : (
+        <div>
+          {this.renderStatusAlert()}
+          <Table
+            columns={Object.keys(this.columns).map(columnKey => ({
+              ...this.columns[columnKey],
+              onSort: () => this.onSortClick(columnKey),
+            }))}
+            data={this.addSupplementalTableElements(this.props.assetsList)}
+            tableSortable
+            defaultSortedColumn="date_added"
+            defaultSortDirection="desc"
+          />
+          {this.renderModal()}
+        </div>
+      );
   }
 }
 
@@ -256,6 +262,16 @@ AssetsTable.propTypes = {
     response: PropTypes.object,
     type: PropTypes.string,
   }).isRequired,
+  courseDetails: PropTypes.shape({
+    lang: PropTypes.string,
+    url_name: PropTypes.string,
+    name: PropTypes.string,
+    display_course_number: PropTypes.string,
+    num: PropTypes.string,
+    org: PropTypes.string,
+    id: PropTypes.string,
+    revision: PropTypes.string,
+  }).isRequired,
   deleteAsset: PropTypes.func.isRequired,
   updateSort: PropTypes.func.isRequired,
   clearAssetsStatus: PropTypes.func.isRequired,
@@ -265,12 +281,14 @@ const mapStateToProps = state => ({
   assetsList: state.assets.list,
   assetsParameters: state.assets.parameters,
   assetsStatus: state.assets.status,
+  courseDetails: state.courseDetails,
 });
 
 const mapDispatchToProps = dispatch => ({
-  deleteAsset: (assetsParameters, assetId) => dispatch(deleteAsset(assetsParameters, assetId)),
-  updateSort: (sortKey, sortDirection) => dispatch(sortUpdate(sortKey, sortDirection)),
   clearAssetsStatus: () => dispatch(clearAssetsStatus()),
+  deleteAsset: (assetsParameters, assetId, courseDetails) =>
+    dispatch(deleteAsset(assetsParameters, assetId, courseDetails)),
+  updateSort: (sortKey, sortDirection) => dispatch(sortUpdate(sortKey, sortDirection)),
 });
 
 const WrappedAssetsTable = connect(
--- a/src/AssetsPage/AssetsTable/index.jsx
+++ b/src/AssetsPage/AssetsTable/index.jsx
@@ -146,9 +146,7 @@ export class AssetsTable extends React.Component {
   deleteAsset() {
     const deletedAsset = { ...this.state.assetToDelete };
 
-    this.props.deleteAsset(this.props.assetsParameters,
-      this.state.assetToDelete.id,
-      this.props.courseDetails);
+    this.props.deleteAsset(this.state.assetToDelete.id, this.props.courseDetails);
 
     this.setState({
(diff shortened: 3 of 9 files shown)


dsjen:
I totally get that this is a WIP, but just to make sure it's on your radar, the `courseId`, https://github.com/edx/studio-frontend/blob/0a4ef1d6b0d81d359dd158c267ac3721ecd21b82/src/data/reducers/assets.js#L6, should probably be what's passed to SFE too!

public/index.html
```diff
@@ -2,6 +2,23 @@
...
+  url_name: "course",
+  name: "edX Demonstration Course",
+  source_of_data: "edx-platform/cms/templates/asset_index.html",
+  display_course_number: "",
+  commit_sha: "30ea0890f4",
+  num: "DemoX",
+  org: "edX",
+  id: "course-v1:edX+DemoX+Demo_Course",
+  revision: "",
+};
```
fysheets: This is set up with the default course settings to be able to test within the SFE container locally

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -111,7 +111,9 @@ export class AssetsTable extends React.Component {
   }
 
   deleteAsset() {
-    this.props.deleteAsset(this.props.assetsParameters, this.state.assetToDelete.id);
+    this.props.deleteAsset(this.props.assetsParameters,
+      this.state.assetToDelete.id,
+      this.props.courseDetails);
```
fysheets: This line was too long for the linter - I'm not sure how I feel about this formatting either, any thoughts?

src/data/actions/assets.js
```diff
@@ -6,9 +6,9 @@ export const requestAssetsSuccess = response => ({
   data: response.assets,
 });
 
-export const getAssets = assetsParameters =>
+export const getAssets = (assetsParameters, courseDetails) =>
   dispatch =>
-    clientApi.requestAssets(assetsParameters.courseId, {
+    clientApi.requestAssets(courseDetails.id, {
```
fysheets: @dsjen updated the API calls to use the ID from courseDetails instead of the hard coded value :)

src/data/reducers/index.js
```diff
@@ -3,9 +3,14 @@ import { combineReducers } from 'redux';
 import assets from './assets';
 import connectionStatus from './connectionStatus';
 
+/* eslint-disable no-undef */
+/* eslint-disable arrow-body-style */
+const courseDetails = () => { return courseContext || {}; };
```
fysheets: Should we make the default here something more than an empty object? For instance, defaults for all of the different values that would be passed on?

src/AssetsPage/AssetsTable/index.jsx
dsjen: Multi line is so hard! I’m ok with what you have here.

src/data/actions/assets.js
dsjen: w00t!

dsjen: w00t!

src/data/reducers/index.js
dsjen: Maybe not even defaulting? You could throw an error if not set?

dsjen: Maybe not even defaulting? You could throw an error if not set?

fysheets: I added the || empty object due to an error I was seeing with redux when the application was starting up because of the `dispatch` in AssetsPage setup:

`dispatch => ({
    getAssets: (assetsParameters, courseDetails) =>
      dispatch(getAssets(assetsParameters, courseDetails)),
  }),`

Let me test again to validate if that error is still occurring if I remove the default.

fysheets: @dsjen - removing the default no longer errors out :) Must have been a byproduct of previous logic. Is there any semi-standard way edX handles throwing/catching errors in javascript?

dsjen: You are paving ways! I think we can throw something like, "Course details required" or something like that.

src/data/reducers/index.js
```diff
@@ -3,9 +3,18 @@ import { combineReducers } from 'redux';
 import assets from './assets';
 import connectionStatus from './connectionStatus';
 
+/* eslint-disable no-undef */
+const courseDetails = () => {
+  if (!courseContext) {
```
dsjen: Nit: it might be clearer to check explicitly if this is undefined.

src/data/reducers/index.js
```diff
@@ -3,9 +3,18 @@ import { combineReducers } from 'redux';
...
+    throw new Error('Error: Missing Course Context');
```
dsjen: Is this test-able?

fysheets: Will try to get some tests added using the reference that @MichaelRoytman gave! http://redux.js.org/docs/recipes/WritingTests.html

src/data/store.test.jsx
```diff
@@ -0,0 +1,14 @@
...
+describe('API client requestAssets', () => {
+  it('throws error when courseDetails is undefined', () => {
+    expect(store.rootReducer.courseDetails.toThrow('Error: Missing Course Context'));
+  });
+
+  it('sets courseDetails properly when defined', () => {
+    courseContext = {};
+    expect(store.rootReducer.courseDetails).toEqual(courseContext);
+  });
+});
```
fysheets: I'm able to get to the courseDetails reducer, but it's giving the following error and the `let courseContext` above doesn't work :(

 `FAIL  src/data/store.test.jsx
  ● Test suite failed to run

    ReferenceError: courseContext is not defined
      
      at courseDetails (src/data/reducers/index.js:8:3)
      at node_modules/redux/lib/combineReducers.js:53:24
          at Array.forEach (<anonymous>)
      at assertReducerShape (node_modules/redux/lib/combineReducers.js:51:25)
      at combineReducers (node_modules/redux/lib/combineReducers.js:107:5)
      at Object.<anonymous> (src/data/reducers/index.js:14:122)
      at Object.<anonymous> (src/data/store.js:5:43)
      at Object.<anonymous> (src/data/store.test.jsx:1:116)
          at Generator.next (<anonymous>)
          at Promise (<anonymous>)
          at Generator.next (<anonymous>)
          at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:188:7)`

Need to investigate the best way to reach that global.

FYI @dsjen

dsjen: My guess is that the `let courseContext` is scoping it to this document only. What if you changed it to a `var` instead?

Also, there might be ways to mock it?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -188,7 +190,10 @@ export class AssetsTable extends React.Component {
   }
 
   render() {
-    return (!this.props.assetsList.length) ? (
+    // TODO: Add UI for when there is nothing in the list and we have a status returned from the API
+    // TODO: http://fhtwd0.axshare.com/#g=1&p=files-and-uploads-empty
```
fysheets: Added these as a reminder to see if we have a ticket created already in the backlog to handle this view.  I talked to Shelby briefly and I don't believe this is needed for feature parity from that discussion, but I didn't' want to forget about it.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -226,24 +228,28 @@ export class AssetsTable extends React.Component {
...
-          data={this.addSupplementalTableElements(this.props.assetsList)}
-          tableSortable
-          defaultSortedColumn="date_added"
-          defaultSortDirection="desc"
-        />
-        {this.renderModal()}
-      </div>
-    );
+    // TODO: Add UI for when there is nothing in the list and we have a status returned from the API
+    // TODO: http://fhtwd0.axshare.com/#g=1&p=files-and-uploads-empty
```
fysheets: These notes are for me to remember to check if we have a ticket set up for this case yet or not. Quick talk with Shelby makes it seem like this might not be necessary for feature parity, but good to keep track of for now.

src/data/reducers/assets.js
```diff
@@ -17,6 +16,8 @@ const list = (state = [], action) => {
       return action.data;
     case assetActions.DELETE_ASSET_SUCCESS:
       return state.filter(asset => asset.id !== action.assetId);
+    case assetActions.ASSET_XHR_FAILURE:
+      return [];
```
fysheets: Reset to the default if we get an error from the API

src/data/store.test.jsx
```diff
@@ -0,0 +1,8 @@
+import store from './store';
+
+describe('Default store values', () => {
```
fysheets: The Jest global is defaulted to `{}` in `package.json`

src/AssetsPage/index.jsx
```diff
@@ -16,13 +16,18 @@ class AssetsPage extends React.Component {
...
+    if (prevProps.courseDetails !== this.props.courseDetails) {
+      this.props.getAssets(this.props.assetsParameters, this.props.courseDetails);
+    }
+
     if (prevProps.assetsParameters !== this.props.assetsParameters) {
+      // if filters changed, update the assetsList
       // TODO: consider using the reselect library for this
-      this.props.getAssets(this.props.assetsParameters);
+      this.props.getAssets(this.props.assetsParameters, this.props.courseDetails);
     }
```
fysheets: Reading back through this, I could see these two combining to be an `||`, I can't remember at the moment if there was a reason for not doing that intentionally.  Can anyone think of a reason why combining these wouldn't be ideal?

package.json
```diff
@@ -72,6 +72,9 @@
     ],
     "transformIgnorePatterns": [
       "/node_modules/(?!(@edx/paragon)/).*/"
-    ]
+    ],
+    "globals": {
+      "courseContext": {}
+    }
```
fysheets: This is used for the reducer tests for `courseDetails`

public/index.html
fysheets: I can easily add it into this PR along with the comment -- should the `base_url` be the localhost? I.e `http://localhost:18010`?

src/AssetsPage/index.jsx
fysheets: 👍 cool I'll `||` these two comparisons then

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -226,24 +228,28 @@ export class AssetsTable extends React.Component {
...
+              onSort: () => this.onSortClick(columnKey),
+            }))}
+            data={this.addSupplementalTableElements(this.props.assetsList)}
+            tableSortable
+            defaultSortedColumn="date_added"
+            defaultSortDirection="desc"
+          />
+          {this.renderModal()}
+        </div>
+      );
```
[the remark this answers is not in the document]
fysheets: I'm okay changing this to the `if/else` because we'll have other cases of what to render later that will also make this harder to read -- i.e. what to render when the list of assets that comes back is empty (because nothing was returned or because there was an error).  I can set that block that is being rendered to a variable and still keep the one return statement and return that variable at the end.

src/data/actions/assets.js
```diff
@@ -34,14 +47,9 @@ export const deleteAssetSuccess = (assetId, response) => ({
   response,
 });
 
-export const assetXHRFailure = response => ({
-  type: assetActions.ASSET_XHR_FAILURE,
-  response,
-});
-
-export const deleteAsset = (assetsParameters, assetId) =>
+export const deleteAsset = (assetsParameters, assetId, courseDetails) =>
```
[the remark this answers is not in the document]
fysheets: Good spot -- I'll remove it and sanity check that nothing breaks.

public/index.html
```diff
@@ -2,6 +2,26 @@
...
+var courseContext = {
+  lang: "fr",
+  url_name: "course",
+  name: "edX Demonstration Course",
+  display_course_number: "",
+  num: "DemoX",
+  org: "edX",
+  id: "course-v1:edX+DemoX+Demo_Course",
+  revision: "",
+  base_url: "http://localhost:18010",
```
fysheets: FYI @MichaelRoytman @efischer19

src/AssetsPage/AssetsTable/index.jsx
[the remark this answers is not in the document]
fysheets: @efischer19 Probably easier to read in the full version of this file than the PR view, but is this easier to read overall?

src/data/actions/assets.js
```diff
@@ -34,14 +47,9 @@ export const deleteAssetSuccess = (assetId, response) => ({
   response,
 });
 
-export const assetXHRFailure = response => ({
-  type: assetActions.ASSET_XHR_FAILURE,
-  response,
-});
-
-export const deleteAsset = (assetsParameters, assetId) =>
+export const deleteAsset = (assetId, courseDetails) =>
```
fysheets: Removed assetParameters as it was no longer being used by this function.  Updated usages across files to handle the function change.

CC @efischer19

src/AssetsPage/AssetsTable/index.jsx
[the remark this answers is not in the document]
efischer19: Looks great!

PR #47, opened 2017-10-27
Added polyfills for fetch and promise.

dsjen:
[EDUCATOR-1604](https://openedx.atlassian.net/browse/EDUCATOR-1604)

I *think* this is what we need to do in order to polyfill fetch, which requires a promise polyfil, fulfilled by using the babel-polyfill (used in [insights](https://github.com/edx/edx-analytics-dashboard/commit/58334e787f57f700dfe5b576b75515df4cac421a) and [marketing](https://github.com/edx/edx-mktg/search?utf8=%E2%9C%93&q=babel-polyfill&type=)).

The reason I included `babel-polyfill` in `index.jsx` is because it's likely of general use (e.g. WeakMap, Array.from, Object.assign).

@edx/educator-dahlia

### DIFF
--- a/src/api/client.js
+++ b/src/api/client.js
@@ -1,5 +1,6 @@
 /* eslint import/prefer-default-export: "off" */
 import Cookies from 'js-cookie';
+import 'whatwg-fetch';  // fetch polyfill
 
 import endpoints from './endpoints';
 
--- a/src/api/client.js
+++ b/src/api/client.js
@@ -1,6 +1,6 @@
 /* eslint import/prefer-default-export: "off" */
 import Cookies from 'js-cookie';
-import 'whatwg-fetch';  // fetch polyfill
+import 'whatwg-fetch'; // fetch polyfill
 
 import endpoints from './endpoints';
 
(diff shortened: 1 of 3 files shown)


src/api/client.js
```diff
@@ -1,5 +1,6 @@
 /* eslint import/prefer-default-export: "off" */
 import Cookies from 'js-cookie';
+import 'whatwg-fetch';  // fetch polyfill
```
dsjen: Same as what's done in marketing, https://github.com/edx/edx-mktg/blob/master/docroot/sites/all/themes/atedx/scripts/react/data/actions/autoSuggest.jsx#L1

src/index.jsx
```diff
@@ -1,3 +1,4 @@
+import 'babel-polyfill';  // general ES2015 polyfill (e.g. promise)
```
dsjen: This is basically what is done in Insights, https://github.com/edx/edx-analytics-dashboard/commit/58334e787f57f700dfe5b576b75515df4cac421a.

dsjen:
Sounds good!

PR #48, opened 2017-11-01
Mark "Delete Asset" header for hiding

efischer19:
# [EDUCATOR-1535](https://openedx.atlassian.net/browse/EDUCATOR-1535)

Depends on https://github.com/edx/paragon/pull/59

PR #49, opened 2017-11-02
mroytman/copy buttons

MichaelRoytman:


MichaelRoytman:
@edx/educator-dahlia This should be good for a review! Please and thank you!

__mocks__/copy-to-clipboard.js
```diff
@@ -0,0 +1 @@
+export default jest.genMockFromModule('copy-to-clipboard');
```
MichaelRoytman: We have to mock the copy-to-clipboard module, since it runs a command that causes all kind of problems for Jest. In any case, we shouldn't be testing 3rd party libraries, so it'll be better to have the copy functionality do nothing.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -4,23 +4,34 @@ import { AssetsTable } from './index';
 
 import { assetActions } from '../../data/constants/actionTypes';
 
+const studioUrl = 'animal';
```
MichaelRoytman: These are used by the copy url buttons.

src/AssetsPage/CopyButton/index.jsx
```diff
@@ -0,0 +1,39 @@
+import React from 'react';
```
MichaelRoytman: I've moved the copy button out into a separate component file. I put it in the AssetsPage directory, because that's where it's being used, but I imagine we could move it to some kind of utility component folder when the time comes and we find that we need this sort of button elsewhere.

src/AssetsPage/CopyButton/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+    this.state = {
+      wasClicked: false,
+    };
+  }
+
+  render() {
+    return (
+      <Button
+        label={!this.state.wasClicked ? this.props.label : this.props.onCopyLabel}
```
MichaelRoytman: The label of the button dynamically switches between the label text and the onCopyLabel text, based on the state of wasClicked. This allows us to display a "Copied!" message (or any other message) inside the actual button, which meets accessibility criteria, allows us not to have to deal with the accessibility issues that accompany a tool-tip, and also affords us more clarity in terms of which URL was copied (compare with a general status alert at the top of the page, which isn't tied to a particular url).

src/AssetsPage/CopyButton/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+        onClick={() => {
+          this.setState({
+            wasClicked: true,
+          });
+          copy(`${this.props.textToCopy}`);
```
MichaelRoytman: This is the actual copy command.

src/AssetsPage/CopyButton/index.jsx
```diff
@@ -0,0 +1,39 @@
...
+        }}
+        onBlur={() => {
```
MichaelRoytman: When focus changes, change the Button label back to the original text.

src/AssetsPage/CopyButton/CopyButton.test.jsx
```diff
@@ -0,0 +1,57 @@
...
+  onCopyLabel: 'Copied!',
+  textToCopy: 'This is my copy text!',
+};
+
+describe('<CopyButton />', () => {
+  let wrapper;
+  let button;
+
+  beforeEach(() => {
+    jest.mock('copy-to-clipboard');
```
MichaelRoytman: I wasn't able to find a way to both mock the module and be able to see how often the function is being called. It would be nice to verify that the CopyButton calls the copy function once per click, but I need to look into how to do that with a manual mock, since I'm only familiar with doing that with mocked functions.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -101,13 +108,29 @@ export class AssetsTable extends React.Component {
...
+    />);
+  }
+
   addSupplementalTableElements() {
     const newAssetsList = this.props.assetsList.map((asset, index) => {
       const currentAsset = Object.assign({}, asset);
       const deleteButton = (<Button
         className={[FontAwesomeStyles.fa, FontAwesomeStyles['fa-trash']]}
         label={''}
-        buttonType={'light'}
```
MichaelRoytman: The focus is really hard to see with this button type, so I've removed it. The default button type has a much clearer outline around it. I think, for the time being, it'll be easier to verify focus this way.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -116,6 +139,9 @@ export class AssetsTable extends React.Component {
       // TODO: create lockButton
       currentAsset.delete_asset = deleteButton;
       // TODO: currentAsset.lock_asset = lockButton;
+
+      currentAsset.urls = this.getCopyUrlButtons(currentAsset.url, currentAsset.external_url);
```
MichaelRoytman: I'm assuming that each asset has a url and an external url. Is that reasonable? Conceptually, I don't see why it would ever be null. But if not, we should discuss what to do in the case that the attribute is null. Do we not render a button? If both are null, is the cell just empty?

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
efischer19: what's the difference between the two?

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -39,11 +40,17 @@ export class AssetsTable extends React.Component {
         key: 'date_added',
         columnSortable: true,
       },
+      urls: {
+        label: 'URLs',
+        key: 'urls',
+        columnSortable: false,
+      },
```
efischer19: Note to @efischer19 - this is going to break the test added in https://github.com/edx/studio-frontend/pull/48 because you used brittle hard-coded expected values. Figure out a smarter way to write that test.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -39,11 +40,17 @@ export class AssetsTable extends React.Component {
...
       delete_asset: {
         label: 'Delete Asset',
         key: 'delete_asset',
         columnSortable: false,
       },
+
```
efischer19: random newline

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -101,13 +108,29 @@ export class AssetsTable extends React.Component {
...
+    );
+  }
+
+  getCopyUrlButton(url, label) {
+    return (<CopyButton
+      label={label}
+      onCopyLabel="Copied!"
+      textToCopy={url}
+    />);
+  }
```
efischer19: Oh that's right, there are 2 different URLs to copy. I've answered my earlier question now

src/AssetsPage/AssetsTable/index.jsx
efischer19: I think it's reasonable to assume each asset has both urls, then not render anything if that assumption fails and the attribute is null

src/AssetsPage/CopyButton/index.jsx
efischer19: Good workaround! The tooltips had me very scared when we were grooming this ticket.

cptvitamin:
This functionality works very well across most OS/Browser/Screen Readers tested.  IE interrupts the user to ask their permission to access the clipboard, and forcibly moves focus to the top of the page.  This interrupts the aria-live message informing the user that the text was copied to the clipboard.  For some unknown reason JAWS 17 in Firefox does not announce the live region update :(  Despite these edge case bugs (not within our control) I am giving this a 👍 and say it's ready to merge.

src/AssetsPage/CopyButton/index.jsx
dsjen: I think it could go in the `src/components` actually. Thoughts?

src/AssetsPage/CopyButton/CopyButton.scss
```diff
@@ -0,0 +1,4 @@
+.button-width {
```
dsjen: I think this is about more than just the button width. Perhaps `copy-button` would be clearer?

src/AssetsPage/CopyButton/index.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+    /*
+      The above copy command removes focus from the button, so we
+      have to explicitly set it back. The copy command works by
+      inserting a span containing the text to be copied into the
+      DOM and running document.execCommand('copy') on it. Some
+      part of this process shifts focus.
+    */
+    this.buttonRef.focus();
+    this.props.onEvent(true);
```
dsjen: I'm finding `onEvent` to be a bit too general, since it's very much indicating the state of the button. I get that there can't be two `onClick` functions though. I honestly can't think of a better name though. Maybe `onCopyFocus` or `onCopyButtonClick`?

src/AssetsPage/CopyButton/CopyButton.scss
MichaelRoytman: Sounds good to me!

src/AssetsPage/CopyButton/index.jsx
MichaelRoytman: Yeah, I find this part of the code kind of... un-indeal, in general. I think the difficulty in naming it reflects that. It's sort of a function that's responsible for both onClick and onBlur on the CopyButton. Is it okay if I go with `onCopyButtonClick` and then explain some of the ambiguity in a comment? I'm totally open to a different structure, but this was sort of necessitated by the need for a separate span on the page to handle screenreader text.

dsjen: Ok, sounds good! `onCopyButtonClick` makes sense (indicates that it's specific to this component) and a comment sounds good given the use case. Thank you!

PR #50, opened 2017-11-03
mroytman/image preview

MichaelRoytman:


MichaelRoytman:
TODO: Use Eric's work to hide column heading for image preview.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -74,7 +81,31 @@ describe('<AssetsTable />', () => {
       expect(wrapper.find('tr')).toHaveLength(defaultProps.assetsList.length + 1);
     });
     it('correct number of trashcan buttons', () => {
-      expect(wrapper.find('button').filterWhere(button => button.hasClass('fa-trash'))).toHaveLength(defaultProps.assetsList.length);
+      expect(wrapper.find('tr button').filterWhere(button => button.hasClass('fa-trash'))).toHaveLength(defaultProps.assetsList.length);
```
MichaelRoytman: I thought it might be good to specify that this trash button is in a table row and not just floating around the page.

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -74,7 +81,31 @@ describe('<AssetsTable />', () => {
...
+    });
+    it('correct number of images', () => {
+      const images = wrapper.find('tr img');
+      expect(images).toHaveLength(defaultProps.assetsList.length);
+    });
+    it('correct type of images', () => {
+      const rows = wrapper.find('tr').filterWhere(row => row.find('td').exists());
+
+      expect(wrapper.find('td img')).toHaveLength(defaultProps.assetsList.length);
```
MichaelRoytman: I want to verify that the images are inside table cells.

MichaelRoytman: Wait. I'm not sure why I have this actually. I'll probably remove it.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -116,6 +125,16 @@ export class AssetsTable extends React.Component {
       // TODO: create lockButton
       currentAsset.delete_asset = deleteButton;
       // TODO: currentAsset.lock_asset = lockButton;
+
+
```
MichaelRoytman: TODO: remove space.

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -101,6 +106,10 @@ export class AssetsTable extends React.Component {
     return this.trashcanRefs[focusAsset.id];
   }
 
+  getImageThumbnail(thumbnail) {
+    return thumbnail ? (<img src={`${this.props.courseDetails.base_url}${thumbnail}`} alt="Description not available" />) : (<img src={undefined} alt="Preview not available" />);
```
MichaelRoytman: I'm kind of surprised the linter isn't complaining about this line. It's definitely longer than 100 characters...

MichaelRoytman:
@edx/educator-dahlia This should be good for a review! Please and thank you!

src/AssetsPage/AssetsTable/index.jsx
dsjen: Nit (well, more like an vague opinion): it might be clearer to define `src` and `alt` outside of the markup and then return the `` tag. e.g.:

```
if (thumbnail === undefined) {
   alt = ...
   src = ...
} else {
   ...
}
return (<img ... />);
```

Maybe too verbose though? It's up to you!

src/AssetsPage/AssetsTable/AssetsTable.test.jsx
```diff
@@ -74,7 +81,31 @@ describe('<AssetsTable />', () => {
...
+
+      const baseUrl = defaultProps.courseDetails.base_url;
```
dsjen: This field is set within SFE itself, but it's not set in Studio, right? Does this need to default to an empty string so that it works when embedded in Studio?

fysheets:
When I test this locally in Studio (http://localhost:18010/assets/course-v1:edX+DemoX+Demo_Course/), the images don't load because it's not grabbing the host URL to fetch the thumbnails.  Shouldn't this be using the host URL when it's not getting it from the Studio pass through `courseDetails`?

`Description not available`

screen shot 2017-11-06 at 11 13 35 am

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -109,7 +109,8 @@ export class AssetsTable extends React.Component {
   }
 
   getImageThumbnail(thumbnail) {
-    return thumbnail ? (<img src={`${this.props.courseDetails.base_url}${thumbnail}`} alt="Description not available" />) : 'Preview not available';
+    const baseUrl = this.props.courseDetails.base_url ? this.props.courseDetails.base_url : '';
+    return thumbnail ? (<img src={`${baseUrl}${thumbnail}`} alt="Description not available" />) : 'Preview not available';
   }
```
fysheets: Thumbnails show up in local studio now! 
screen shot 2017-11-07 at 10 25 50 am

src/AssetsPage/AssetsTable/index.jsx
```diff
@@ -102,6 +108,11 @@ export class AssetsTable extends React.Component {
     return this.trashcanRefs[focusAsset.id];
   }
 
+  getImageThumbnail(thumbnail) {
+    const baseUrl = this.props.courseDetails.base_url ? this.props.courseDetails.base_url : '';
```
dsjen: Nit: What do you think about using [default props](https://reactjs.org/docs/react-component.html#defaultprops) instead of the ternary?

PR #51, opened 2017-11-06
Assets Page Layout

fysheets:


config/webpack.dev.config.js
```diff
@@ -44,7 +44,7 @@ module.exports = Merge.smart(commonConfig, {
             options: {
               sourceMap: true,
               modules: true,
-              localIdentName: '[path][name]__[local]--[hash:base64:5]',
```
fysheets: @arizzitano @edx/educator-dahlia Do we want to rename our existing scss files to be componentNameStyles.scss?  For example: `AssetsPageStyles.scss` instead of `AssetsPage (directory)` > `styles.scess` -- if so, I think we should be able to update webpack here to be `[name]__[local]` and still have the unique name.

dsjen: Actually, Now that you write it out, I sort of like that it's `style.scss`, which mimics `index.jsx` per directory/component.

fysheets: I can do that for the moment -- the idea for making it a more discrete name is to change the style name from being something like `src-AssetsPage-AssetsFilters-styles__filter-set` to `AssetsFiltersStyles__filter-set`

fysheets:
Test functionality at: https://studio-fysheets.sandbox.edx.org/assets/course-v1:edX+DemoX+Demo_Course/

src/AssetsPage/AssetsPage.scss
```diff
@@ -0,0 +1,3 @@
+.assets {
+	
+}
```
fysheets: I left this in place because I'm assuming we will need some overarching page styles and that this will be used soon anyhow (it used to house the pink).

src/AssetsPage/index.jsx
```diff
@@ -38,8 +39,16 @@ class AssetsPage extends React.Component {
...
+        <div className={edxBootstrap.container}>
+          <div className={edxBootstrap.row}>
+            <div className={edxBootstrap.col}>
+              <WrappedAssetsFilters />
+            </div>
+            <div className={edxBootstrap['col-10']}>
+              <WrappedAssetsTable />
+            </div>
+          </div>
+        </div>
```
fysheets: Bootstrap grids are based off of a 12 column grid

PR #53, opened 2017-11-14
Added lock asset button.

dsjen:
[EDUCATOR-1534](https://openedx.atlassian.net/browse/EDUCATOR-1534)

This adds asset locking to assets in the asset table.

@edx/educator-dahlia - ~~I'm currently adding tests, but feel free to give early feedback as the functionality should be working.~~ (tests have been added)


src/data/reducers/assets.js
```diff
@@ -10,13 +11,45 @@ const initialState = {
   direction: 'desc',
 };
 
+const toggleLockAsset = (assets, assetId) => assets.map(asset => (
+  asset.id === assetId ? { ...asset, locked: !asset.locked } : asset));
```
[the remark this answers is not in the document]
MichaelRoytman: Assuming that you need to flip the value because the API value is the opposite of what you want, this is a super simple example of what we could do in a reselect selector. We could keep the state as-is here and compute the value that is necessary for the table in a selector. I don't think this case is worth the hassle of setting it up, but perhaps it gives us some intuition of when we could use it?

src/data/constants/loadingTypes.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/prefer-default-export */
+export const assetLoading = {
+  LOCK: 'LOCK',
```
MichaelRoytman: Is this where other kinds of loading would go?

Like

```
export const tableLoading = {
   FILTER: 'FILTER',
   SORT: 'SORT',
   ...
}
```

and

```
export const assetLoading = {
   DELETE: 'DELETE'
   ...
}
```

?

src/data/reducers/assets.js
```diff
@@ -32,8 +65,14 @@ const status = (state = {}, action) => {
         response: action.response,
         type: action.type,
       };
-    case assetActions.ASSET_XHR_FAILURE:
+    case assetActions.DELETE_ASSET_FAILURE:
```
MichaelRoytman: This is sort of what makes me think that maybe a generic XHRFailure would be useful, because the only difference between the DELETE_ASSET_FAILURE and TOGGLING_LOCK_ASSET_FAILURE are that the TOGGLING_LOCK_ASSET_FAILURE gets the asset it's trying to lock (I'd argue that DELETE_ASSET_FAILURE could use that too). So generalizing it to be any ASSET_XHR_FAILURE would reduce the redundancy here, perhaps.

src/components/AssetsTable/index.jsx
```diff
@@ -108,9 +142,43 @@ export class AssetsTable extends React.Component {
     return this.trashcanRefs[focusAsset.id];
   }
 
-  getImageThumbnail(thumbnail) {
-    const baseUrl = this.props.courseDetails.base_url ? this.props.courseDetails.base_url : '';
-    return thumbnail ? (<img src={`${baseUrl}${thumbnail}`} alt="Description not available" />) : 'Preview not available';
+  getUpdatingLockButton(asset) {
```
MichaelRoytman: Nit: Could we, maybe, choose a consistent name for the concept of "loading?" Either toggling, updating, loading, requesting, fetching, whatever. I'm finding I have to kind of... translate them all to the same concept, and having a single word for the concept of "updating" would be easier, I think. I think "loading" is a good catch-all terms for various kinds of requests (e.g. can't toggle a delete), but up to you!

src/data/actions/assets.js
```diff
@@ -54,9 +53,39 @@ export const deleteAsset = (assetId, courseDetails) =>
...
+  asset,
+});
+
+export const toggleLockAssetFailure = (asset, response) => ({
+  type: assetActions.TOGGLING_LOCK_ASSET_FAILURE,
+  asset,
+  response,
+});
+
+export const toggleLockAsset = (asset, courseDetails) =>
```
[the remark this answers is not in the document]
MichaelRoytman: I think the discussion at the top of this page would be useful. https://redux.js.org/docs/advanced/AsyncActions.html I think we basically just need to come to a decision about how we want to do it and document it somewhere so that we do it consistently.

src/components/AssetsTable/index.jsx
```diff
@@ -88,6 +95,33 @@ export class AssetsTable extends React.Component {
...
+  getLockButton(asset) {
+    const classes = [FontAwesomeStyles.fa];
+    let lockState;
+    if (asset.locked) {
+      lockState = 'Locked';
+      classes.push(FontAwesomeStyles['fa-lock']);
+    } else {
+      lockState = 'Unlocked';
+      classes.push(FontAwesomeStyles['fa-unlock']);
+    }
```
MichaelRoytman: The pushing is neat!

src/data/reducers/assets.js
```diff
@@ -10,13 +11,45 @@ const initialState = {
...
+
 const list = (state = [], action) => {
+  let assets;
   switch (action.type) {
     case assetActions.REQUEST_ASSETS_SUCCESS:
       return action.data;
     case assetActions.DELETE_ASSET_SUCCESS:
       return state.filter(asset => asset.id !== action.assetId);
-    case assetActions.ASSET_XHR_FAILURE:
+    case assetActions.TOGGLE_LOCK_ASSET_SUCCESS:
```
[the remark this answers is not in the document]
MichaelRoytman: I think what's throwing me off, maybe, is that we have a dedicated status piece of Redux state, which represents the general status of requests made that are relevant to the table, right? It just seems to me like the status of a request should similarly be separate from the actual asset. I realize that the request is made ON the asset, so it makes sense to have it there, but, on the other hand, the status Redux state currently consumes the status of a delete request, which is also on the asset. Am I making sense?

src/data/actions/assets.js
```diff
@@ -6,8 +6,8 @@ export const requestAssetsSuccess = response => ({
   data: response.assets,
 });
 
-export const assetXHRFailure = response => ({
-  type: assetActions.ASSET_XHR_FAILURE,
+export const assetDeleteFailure = response => ({
+  type: assetActions.DELETE_ASSET_FAILURE,
```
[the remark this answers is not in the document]
MichaelRoytman: We may not even want to do this, by the way. It's just an idea that would prevent us from having to write the same kind of boilerplate code over and over.

MichaelRoytman:
P.S. Since this is the first PR to tackle loading states (blazing trails 🔥 ), I wanted to plop these resources down here, both of which are authored by Dan Abramov.

[StackOverflow Answer by Dan on Loading Indicators](https://stackoverflow.com/questions/35456935/how-to-show-a-loading-indicator-in-react-redux-app-while-fetching-the-data/35643128#35643128)
[Egghead Tutorial on Loading Indicator](https://egghead.io/lessons/javascript-redux-displaying-loading-indicators)
[Example of isLoading State in Redux, explained by above StackOverflow answer](https://github.com/reactjs/redux/tree/master/examples/async)

MichaelRoytman:
P.P.S. Have you ventured into the scary world of other browsers? Does everything work as expected?

src/components/AssetsTable/index.jsx
```diff
@@ -88,6 +95,33 @@ export class AssetsTable extends React.Component {
     });
   }
 
+  onLockClick(asset) {
+    this.props.toggleLockAsset(asset, this.props.courseDetails);
+  }
+
+  getImageThumbnail(thumbnail) {
+    const baseUrl = this.props.courseDetails.base_url ? this.props.courseDetails.base_url : '';
```
jaebradley: Nit: you could consider `this.props.courseDetails.base_url || ''` instead

src/components/AssetsTable/index.jsx
```diff
@@ -88,6 +95,33 @@ export class AssetsTable extends React.Component {
...
+    return thumbnail ? (<img src={`${baseUrl}${thumbnail}`} alt="Description not available" />) : 'Preview not available';
```
jaebradley: It feels slightly weird to try and get an image thumbnail for an `undefined` object. Especially since the value returned if `thumbnail` is `undefined`is not an image per se, but a `string` value.

Is there any reason why this control logic can't live a level above like

```javascript
if (currentAsset.thumbnail) { 
  currentAsset.image_preview = this.getImageThumbnail(currentAsset.thumbnail) } 
else { 
  currentAsset.image_preview = 'Preview not available' 
}
```

src/components/AssetsTable/index.jsx
```diff
@@ -88,6 +95,33 @@ export class AssetsTable extends React.Component {
...
+      classes.push(FontAwesomeStyles['fa-lock']);
+    } else {
+      lockState = 'Unlocked';
+      classes.push(FontAwesomeStyles['fa-unlock']);
+    }
+    return (<Button
+      label={(<span className={classNames(...classes)} />)}
+      buttonType={'light'}
+      aria-label={`${lockState} ${asset.display_name}`}
+      onClick={() => { this.onLockClick(asset); }}
```
jaebradley: I think this will create a new function instance when React compares component trees meaning that this will always trigger a re-render.

Check out [this article](https://github.com/vasanthk/react-bits/blob/master/gotchas/01.pure-render-checks.md#bad-2).

Here's [a more in-depth article](https://medium.com/@esamatti/react-js-pure-render-performance-anti-pattern-fb88c101332f)
> In Javascript different instances have different identities and thus the shallow equality check always produces false and tells React to re-render the components.

dsjen:
Thanks for the reviews! I've pushed up tests and will begin addressing comments tomorrow.

src/components/AssetsTable/index.jsx
dsjen: yep, cleaner!

src/components/AssetsTable/index.jsx
dsjen: I'm going to punt on this one for now. The only reason this code was marked as new is because I moved some functions around, and there's a bunch of other decision I want to tackle in this PR. Thanks for the suggestion though!

src/data/constants/loadingTypes.js
dsjen: Yep! I'd like to see this become the home for that. What do you think?

src/data/reducers/assets.js
```diff
@@ -10,13 +11,45 @@ const initialState = {
   direction: 'desc',
 };
 
+const toggleLockAsset = (assets, assetId) => assets.map(asset => (
```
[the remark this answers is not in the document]
dsjen: I originally had it stepped down, but get: `44:14  error  'toggleLockAsset' was used before it was defined  no-use-before-define`. I think the issue here is that the const is hoisted (and undefined initially), so the right way is to declare it above the reducer.

I don't think that addresses the confusion you had in this though! I could make a utils in the `reducers` directory perhaps and move these functions into it. What do you think?

src/data/reducers/assets.js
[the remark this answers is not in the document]
dsjen: So, in this case, the asset that is being flipped is the *original* asset. The `PUT`'s response doesn't include the entire asset.

I absolutely agree about evaluating reselect (in the future!).

src/data/reducers/assets.js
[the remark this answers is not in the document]
dsjen: You're making perfect sense! We had this discussion at arch earlier this week (?). We could sure a list of assets that have some sort of "loading" happening on them. e.g. `...status.loading.delete = ['asset-id-1', 'asset-id-2']` and the same for locking. We decided that adding a field on an asset would make more sense as it's part of the table *state* for an asset row.

I'm inclined to keep this as is as there doesn't seem to be a clear winner (right?).

src/data/actions/assets.js
```diff
@@ -54,9 +53,39 @@ export const deleteAsset = (assetId, courseDetails) =>
         if (response.ok) {
           dispatch(deleteAssetSuccess(assetId, response));
         } else {
-          dispatch(assetXHRFailure(response));
+          dispatch(assetDeleteFailure(response));
+        }
+      });
+
+export const togglingLockAsset = (asset, response) => ({
```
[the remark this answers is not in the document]
dsjen: Good catch!

src/data/actions/assets.js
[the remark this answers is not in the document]
dsjen: I wanted to separate this out because regardless, we'd need to differentiate the types of errors. I could potentially see the differentiation localized somewhere in redux (e.g. setting the message, error severity, etc.), but ultimately the frontend does need to know about the specifics, especially when we account for i18n. The text will need to be marked for translation in the view.

Anyway, I think we might be able to achieve all our goals with some tweaking and probably easier to get on the same page in person. Look for our a meeting, Dahlia!

src/components/AssetsTable/index.jsx
```diff
@@ -212,28 +283,22 @@ export class AssetsTable extends React.Component {
...
-        break;
+    const deleteActions = [assetActions.DELETE_ASSET_FAILURE, assetActions.DELETE_ASSET_SUCCESS];
+    let status = {
+      alertDialog: 'Processing',
+      alertType: 'info',
+    };
+    if (assetsStatus.type === assetActions.TOGGLING_LOCK_ASSET_FAILURE) {
+      status = this.getLockFailedStatusFields(assetsStatus);
+    } else if (deleteActions.includes(assetsStatus.type)) {
+      status = this.getDeleteStatusFields(assetsStatus);
```
fysheets: I like how this turned out :)

src/components/AssetsTable/index.jsx
dsjen: Nice catch! Thanks!

dsjen:
@MichaelRoytman -- I just tested it on FF, Safari, and Chrome and it works! Feel free to re-review!

N.B. I'm setting a data attribute on the button that I can then grab in order to address the inline function re-render issue.

PR #54, opened 2017-11-17
Add pagination component to SFE

fysheets:
This will add a styled pagination component to the bottom of the assets page. It currently refactors our reducer states so that each component only gets the pieces of state relevant to it's tasks, instead of possibly rerendering based on unrelated changes in a larger object.

This is still a work in progress.


fysheets:
Left todo: merge in react16 branch, styling, tests

src/data/actions/assets.js
```diff
@@ -1,31 +1,47 @@
 import * as clientApi from '../api/client';
 import { assetActions } from '../constants/actionTypes';
 
+const isCorrectResponse = (request, lastRequest) => (
```
MichaelRoytman: This function verifies whether the current state of lastRequest (the request currently stored in Redux) is equal to the request that was made in the currently executing getAssets call. This accounts for cases where multiple requests are made and the underlying Redux request state may change from underneath an API call and result in inconsistent state in the UI.

An example is clicking all the filters, unclicking each filter one by one (from Other up to Audio) and ending up with only Audio files displayed (instead of all files). This happens because the API call that is made last (all documents) comes back long before the second to last API call (audio).

src/data/reducers/assets.js
```diff
@@ -5,19 +5,44 @@ import { assetLoading } from '../constants/loadingTypes';
 
 import { addLoadingField, removeLoadingField, toggleLockAsset } from './utils';
 
-const initialState = {
+const filtersInitial = {
```
MichaelRoytman: This is the initial state of the filters state.

src/data/reducers/assets.js
```diff
@@ -5,19 +5,44 @@ import { assetLoading } from '../constants/loadingTypes';
...
+  assetTypes: {},
+};
+
+const metaDataInitial = {
```
MichaelRoytman: This is the initial state of the table metadata.

src/data/reducers/assets.js
```diff
@@ -5,19 +5,44 @@ import { assetLoading } from '../constants/loadingTypes';
...
+  start: 0,
+  end: 0,
   page: 0,
   pageSize: 50,
-  assetTypes: {},
+  totalCount: 0,
+};
+
+const sortOrderInitial = {
```
MichaelRoytman: This is the initial state of the table sort order.

src/data/reducers/assets.js
```diff
@@ -5,19 +5,44 @@ import { assetLoading } from '../constants/loadingTypes';
...
   sort: 'date_added',
   direction: 'desc',
 };
 
+const requestInitial = {
```
MichaelRoytman: This is the initial request state of the assets page.

src/data/reducers/assets.js
```diff
@@ -5,19 +5,44 @@ import { assetLoading } from '../constants/loadingTypes';
...
+  ...filtersInitial,
+  ...metaDataInitial,
+  ...sortOrderInitial,
+};
+
+
+const filters = (state = filtersInitial, action) => {
```
MichaelRoytman: Filters is a slice of the Redux state that represents filters on the assets. It represents the actual, current state of the table's filters.

src/data/reducers/assets.js
```diff
@@ -34,6 +59,33 @@ const list = (state = [], action) => {
...
+        page: action.data.page,
+        pageSize: action.data.pageSize,
+        totalCount: action.data.totalCount,
+      };
+    default:
+      return state;
+  }
+};
+
+const sortMetaData = (state = sortOrderInitial, action) => {
```
MichaelRoytman: sortMetaData is a slice of the Redux state that represents sort on the assets. It represents the actual, current state of the table's sorting.

src/data/reducers/assets.js
```diff
@@ -59,21 +111,36 @@ const status = (state = {}, action) => {
...
+        ...state,
+        assetTypes: { ...state.assetTypes, ...action.data },
+      };
     default:
       return state;
   }
 };
 
-const assets = combineReducers({
+export const assets = combineReducers({
```
MichaelRoytman: This is sort of a relic left over from refactoring. I don't know whether it's really necessary to wrap the list in a combineReducers function? Should I leave it or look into just exporting list? Or exporting only list but also renaming it to assets?

src/data/reducers/index.js
```diff
@@ -1,13 +1,15 @@
 import { combineReducers } from 'redux';
 
-import assets from './assets';
+import { assets, metadata, request } from './assets';
```
MichaelRoytman: The changes made to this file is just a reflection of the changes to reducers/assets.js.

src/data/actions/assets.js
```diff
@@ -1,31 +1,47 @@
...
     })
       .then((response) => {
         if (response.ok) {
           return response.json();
         }
         throw new Error(response);
       })
-      .then(json => dispatch(requestAssetsSuccess(json)))
+      .then((json) => {
+        const lastRequest = getState().request;
```
MichaelRoytman: `lastRequest` might not be clear. It's a little difficult to describe what situation this is solving, since both requests are actual requests. lastRequest is the last request that was recorded as necessary to make based on changes to the UI state, whereas request refers to the request in this current getAssets call. Does that make sense?

src/data/actions/assets.js
```diff
@@ -1,31 +1,47 @@
...
 
 export const assetDeleteFailure = response => ({
   type: assetActions.DELETE_ASSET_FAILURE,
   response,
 });
 
-export const getAssets = (assetsParameters, courseDetails) =>
-  dispatch =>
+export const getAssets = (request, courseDetails) =>
+  (dispatch, getState) =>
```
MichaelRoytman: I'm adding in getState as a parameter so that I can access the current Redux store state.

src/components/Pagination/Pagination.test.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+    wrapper.setProps({ updatePage: updatePageSpy });
+
+    pageLinks.last().simulate('click');
+
+    expect(updatePageSpy).toHaveBeenCalledTimes(1);
+    // API treats pages as 0-indexed, but we treat them as 1-indexed
+    expect(updatePageSpy).toHaveBeenCalledWith(totalPages - 1);
+  });
+  it('has correct mapDispatchToProps', () => {
```
MichaelRoytman: We have not been testing mapDispatchToProps, but this is a way to test it that I found recommended online, so it might be useful to add this sort of test to the other components (at a later date).

MichaelRoytman:
@edx/educator-dahlia This should be ready for a review! I made some changes to the Redux store that might make more sense to review first before viewing the diffs of the other components. There was a little bit of refactoring and renaming that might make changes to other components look extra-complicated without looking at the refactoring first.

src/components/Pagination/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+  constructor(props) {
+    super(props);
+    this.onPageClick = this.onPageClick.bind(this);
+  }
+
+  onPageClick(pageData) {
+    this.props.updatePage(pageData.selected);
+  }
+
+  getDisabledScreenReaderText() {
```
MichaelRoytman: This is actually not going to be this simple, since the disabled screen reader text will probably be unique for the "..." "button".

src/components/Pagination/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+    return (
+      <span className={paginationStyles['sr-only']}> button disabled</span>
+    );
+  }
+
+  getPreviousLabel(totalPages) {
+    return (
+      <span>
+      previous
+        {(this.props.assetsListMetaData.page === 0 && totalPages > 0) &&
```
MichaelRoytman: Actually, I believe that if there are 0 pages, the previous button is still marked as disabled by the Pagination component, so I may have to change this to >= 0. I'll have to verify.

src/data/api/client.js
```diff
@@ -3,16 +3,7 @@ import Cookies from 'js-cookie';
...
-/*
-API returns snake-cased attribute names, but API expects names that match database attribute names,
-which are inconsistently cased
-*/
-const assetAPIAttributesToDatabaseAttributes = {
-  display_name: 'displayname',
-  content_type: 'contentType',
-  date_added: 'uploadDate',
-};
+import { getDatabaseAttributesFromAssetAttributes } from '../../utils/getAttributes';
```
MichaelRoytman: I'm pulling this functionality out into a utils file, since I found a need for something very similar elsewhere.

src/utils/getAttributes.js
```diff
@@ -0,0 +1,56 @@
+/*
```
MichaelRoytman: I hope this comment explains the need for these utility functions. Please let me know if it's unclear, and I can clarify.

src/utils/getAttributes.js
```diff
@@ -0,0 +1,56 @@
...
+    "date_added": "Dec 19, 2017 at 16:34 UTC",
+    ...
+  }
+}
+
+Note the snake-cased attribute values for "sort" and the keys of "assets"[0].
+
+*/
+
+const assetAPIAttributesToDatabaseAttributes = {
```
MichaelRoytman: I still hate the names of these functions, but I honestly can't think of better ones 😞 .

src/components/Pagination/Pagination.scss
```diff
@@ -0,0 +1,25 @@
...
+    justify-content: center;
+    align-items: center;
+}
+
+.page-item {
+    min-width: 50px;
+    text-align: center;
+}
+
+.disabled {
```
MichaelRoytman: We're trying to leverage the bootstrap classnames for pagination for the pagination component. The [scss](https://github.com/twbs/bootstrap/blob/v4-dev/scss/_pagination.scss#L45) expects that a disabled pagination "button" has a page-item parent with a page-link child that has the disabled classname on the page-link. However, the pagination package we're using puts the disabled class name on the page-item parent instead of the page-link class. Therefore, we have to create this disabled class that we can use as a classname on the [page-link](https://github.com/edx/studio-frontend/pull/54/files#diff-19113b55aba41492c4d2adb0959b1478R77).

src/components/Pagination/Pagination.scss
```diff
@@ -0,0 +1,25 @@
...
+  color: $pagination-disabled-color;
+  pointer-events: none;
+  background-color: $pagination-disabled-bg;
+  border-color: $pagination-disabled-border-color;
+}
+
+.disabled {
```
MichaelRoytman: The pagination package does not put the disabled classname on the break ("..."), so we're adding this rule to disable the break.

src/data/actions/assets.js
dsjen: I think this make sense. I was trying to think about other words too, such as "most recent" or "current", but "last" seems just fine!

src/data/reducers/assets.js
thallada: It seems like exporting assets as just list should work (`export const assets = list;`)

src/data/reducers/index.js
```diff
@@ -10,6 +10,8 @@ const studioDetails = () => studioContext;
 const rootReducer = combineReducers({
```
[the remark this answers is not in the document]
thallada: I could very well misunderstand how redux works, but I thought you had to have one "reducer" function that redux calls. `combineReducer` allows you to split that into multiple separate functions and still have that single top-level function.

src/utils/getAttributes.js
thallada: This is very clear! I like the examples :100:

src/data/reducers/assets.js
MichaelRoytman: Do you think that that's necessary? I feel like at that point I should just rename the list reducer to be assets. What do you think?

src/data/actions/assets.js
MichaelRoytman: Thanks!

src/data/reducers/assets.js
thallada: Oh true, that does sound better.

src/utils/getAttributes.js
MichaelRoytman: Good point, Dennis. I'll change the name.

src/components/AssetsTable/index.jsx
```diff
@@ -71,6 +71,7 @@ export class AssetsTable extends React.Component {
     this.statusAlertRef = {};
 
     this.onDeleteClick = this.onDeleteClick.bind(this);
+    this.onSortClick = this.onSortClick.bind(this);
```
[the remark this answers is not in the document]
MichaelRoytman: I actually didn't need to bind the function, since it's being passed to the Table and is bound there. I'll remove that line.

src/data/actions/assets.js
MichaelRoytman: I agree. I think Dennis and I had talked about this being a better option, but I don't know how widely supported that is, per [this](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) page. It sounds like not all browsers support aborting a fetch ('experimentally' also doesn't inspire much confidence...), so I'm not sure how tractable that will be if we're using fetch?

src/data/reducers/index.js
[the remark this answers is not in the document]
MichaelRoytman: According to [this](https://redux.js.org/docs/api/combineReducers.html#combinereducersreducers) documentation (second paragraph), you pass a single function to the [createStore](https://redux.js.org/docs/api/createStore.html) function. The combineReducers call here creates a single "root reducer" that you pass to the createStore function, so it is necessary to group together all the reducers in this way. That's the purpose of the combineReducers function - to be able to split and slice your store state. Since they're all sharing the same store, we have to group them together, even if they operate on different parts of the store, I believe.

src/data/actions/assets.js
thallada: Wow, I'm surprised there's no way to cancel a request right now... seems like a pretty basic feature. It looks like that [the fetch pollyfill](https://github.com/edx/studio-frontend/blob/master/package.json#L30) that we are using is going to add support for aborts soon though: https://github.com/github/fetch/pull/592.

thallada: If we really wanted to implement this we should consider dropping fetch for straight up XHR requests because they support cancelling ([but there might be some other drawbacks to them](https://stackoverflow.com/a/35552420/6620612)). I think [axios](https://github.com/axios/axios) is a good XHR client library that we could use.

MichaelRoytman:
Changed lines since last review:
[actions/assets.js](https://github.com/edx/studio-frontend/pull/54/files#diff-a8836bb769c6a2335b5d0ef0ce01457bR4) (line 4-12)
[reducers/assets.js](https://github.com/edx/studio-frontend/pull/54/files#diff-1dcfcb67a72beb3a12a086472ae272aeR42) (renamed list reducer to assets)
[utils/getAssetsAttributes](https://github.com/edx/studio-frontend/pull/54/files#diff-747e0e1f7fbcf7915a7abc94c29ae6abR38) (lines 38-54)

thallada:
Updates look good to me :+1:

PR #55, opened 2017-11-20
Upgrade SFE to React 16

fysheets:
The pagination package in use has a peer dependency of react 16, upgrade our project accordingly.


MichaelRoytman:
@edx/educator-dahlia This should be ready for review!

shim.js
```diff
@@ -0,0 +1,3 @@
+global.requestAnimationFrame = (callback) => {
```
jaebradley: It might be useful adding a `TODO` and / or the link to the issue explaining this workaround (https://github.com/facebook/jest/issues/4545#issuecomment-333004504).

src/components/AssetsFilters/AssetsFilters.test.jsx
```diff
@@ -1,7 +1,12 @@
 import React from 'react';
-import { mount } from 'enzyme';
+import Enzyme from 'enzyme';
+import Adapter from 'enzyme-adapter-react-16';
 import { AssetsFilters } from './index';
 
+Enzyme.configure({ adapter: new Adapter() });
```
jaebradley: I think you only need to configure this once so you can move this `Adapter` setup into a `setupTest.js` file and thus you don't have to call this at the beginning of all of the test files.

src/components/AssetsTable/AssetsTable.test.jsx
```diff
@@ -457,11 +463,11 @@ describe('<AssetsTable />', () => {
     it('moves from modal to trashcan on modal close', () => {
       trashButtons.at(0).simulate('click');
 
-      expect(closeButton.matchesElement(document.activeElement)).toEqual(true);
+      expect(closeButton.html()).toEqual(document.activeElement.outerHTML);
```
dsjen: @fysheets -- oh, that's how you did it! 👍

fysheets: yeah! @jaebradley helped logic this one out :) that `.outerHTML` was the key

jaebradley: I'm not sure this is the best way to do it...but it works #yolo

alt-text

PR #57, opened 2017-11-29
Add Accessibility Policy Page

fysheets:
This creates an additional webpack entry point (index/bundle) for the accessibility page. It also creates a base page that we can add all of the relevant information/styling into.

**Note** This solution is to be able to test the page from within the SFE container, and creating a new index.jsx type of file for each page we want to create is not a long term solution. This is being done with regards to our current timeline.

### DIFF
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -16,12 +16,20 @@ targetUrl = `http://${targetUrl}:18010`;
 
 module.exports = Merge.smart(commonConfig, {
   devtool: 'cheap-module-eval-source-map',
-  entry: [
-    // enable react's custom hot dev client so we get errors reported
-    // in the browser
-    require.resolve('react-dev-utils/webpackHotDevClient'),
-    path.resolve(__dirname, '../src/index.jsx'),
-  ],
+  entry: {
+    assets: [
+      // enable react's custom hot dev client so we get errors reported
+      // in the browser
+      require.resolve('react-dev-utils/webpackHotDevClient'),
+      path.resolve(__dirname, '../src/index.jsx'),
+    ],
+    accessibilityPolicy: [
+      // enable react's custom hot dev client so we get errors reported
+      // in the browser
+      require.resolve('react-dev-utils/webpackHotDevClient'),
+      path.resolve(__dirname, '../src/accessibilityIndex.jsx'),
+    ],
+  },
   module: {
     rules: [
       {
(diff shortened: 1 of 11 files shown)


config/webpack.dev.config.js
```diff
@@ -16,12 +16,20 @@ targetUrl = `http://${targetUrl}:18010`;
...
+      path.resolve(__dirname, '../src/index.jsx'),
+    ],
+    a11yPolicy: [
+      // enable react's custom hot dev client so we get errors reported
+      // in the browser
+      require.resolve('react-dev-utils/webpackHotDevClient'),
+      path.resolve(__dirname, '../src/a11yIndex.jsx'),
+    ],
+  },
   module: {
```
fysheets: For right now, if you want to test with the container, you will need to comment out the assets entry point so that the dev server will point to the accessibility index page instead

fysheets:
screen shot 2017-11-29 at 1 10 57 pm

View of AccessibilityPolicyPage placeholder when the assets entry point is commented out from the dev config.

config/webpack.dev.config.js
arizzitano: *this* would actually be a good reason to implement routing via react-router or something similar. That way, we could keep a single entry point here and not have to worry about switching between pages by commenting things out.

package.json
```diff
@@ -76,7 +76,8 @@
       "/node_modules/(?!(@edx/paragon)/).*/"
     ],
     "globals": {
-      "courseContext": {}
+      "courseContext": {},
```
dsjen: I'm sort of wondering if instead of `courseContext`/`zendeskContext`, we could just have a standard context. Perhaps `studioContext` and this could be the course or zendesk depending on which page is grabbing the data.

This might need some upstream changes in the platform though, but it might make @MichaelRoytman's work for getting studio passing over the zendesk info a bit easier as he could reuse the existing context inserting function that @efischer19 created initially.

fysheets: Totally and could help with scaling. Are you thinking of something like `studioContext: { courseContext: {...}, zendeskContext: {...}}` or something where the data that's currently within `courseContext` would be in this new `studioContext`?

dsjen: I think it `studioContext` could just replace what's in `courseContext` -- i.e. the structure of the content is left entirely up to the page. Just an idea! I think it would be more flexible.

fysheets: Got it! I'll take a look and make sure that changing that doesn't mess with other assumptions made elsewhere in the repo.

config/webpack.dev.config.js
fysheets: I think this is why I had initially defaulted to react-router in my thought process, before realizing the implications to integrating within the rest of studio/prod. I'd be happy to have a sit down (after this push) to create stories to handle our dev/prod in a cleaner way for future development as well. :)

MichaelRoytman:
Can we refer to the accessibility page by accessibility? I think the a11y term just kind of adds an extra layer of abstraction that some people might not be familiar with.

fysheets:
@MichaelRoytman Definitely -- part of the a11y choice was because I wasn't 100% on if we could keep that page how it was anyway.

src/data/store.test.jsx
```diff
@@ -2,7 +2,7 @@ import store from './store';
 
 describe('Default store values', () => {
   it('sets courseDetails properly when defined in global', () => {
```
dsjen: The description should probably be updated too! :)

fysheets: Good catch! Thanks!

PR #58, opened 2017-12-01
Added accessibility policy body component.

dsjen:
This adds a component for displaying the body of the accessibility request page. The phone number needs to be updated when it's settled.

@edx/educator-dahlia

### DIFF
--- a/src/components/AccessibilityBody/AccessibilityBody.test.jsx
+++ b/src/components/AccessibilityBody/AccessibilityBody.test.jsx
@@ -1,0 +1,25 @@
+import React from 'react';
+import Enzyme from 'enzyme';
+import AccessibilityBody from './index';
+
+const mount = Enzyme.mount;
+
+let wrapper;
+
+describe('<AccessibilityBody />', () => {
+  describe('renders', () => {
+    beforeEach(() => {
+      wrapper = mount(
+        <AccessibilityBody
+          communityAccessibilityLink="http://example.com"
+          phoneNumber="555-555-5555"
+          email="example@example.com"
+        />,
+      );
+    });
+    it('contains links', () => {
+      expect(wrapper.text()).toContain('example@example.com');
+      expect(wrapper.text()).toContain('555-555-5555');
+    });
+  });
+});
--- a/src/components/AccessibilityBody/AccessibilityBody.test.jsx
+++ b/src/components/AccessibilityBody/AccessibilityBody.test.jsx
@@ -20,6 +20,7 @@ describe('<AccessibilityBody />', () => {
     it('contains links', () => {
       expect(wrapper.text()).toContain('example@example.com');
       expect(wrapper.text()).toContain('555-555-5555');
+      expect(wrapper.find({ href: 'http://example.com' })).toHaveLength(1);
     });
   });
 });
--- a/src/components/AccessibilityBody/index.jsx
+++ b/src/components/AccessibilityBody/index.jsx
@@ -4,10 +4,8 @@ import PropTypes from 'prop-types';
 const AccessibilityBody = (props) => {
   const mailto = `mailto:${props.email}`;
   const emailElement = (<a href={mailto}>{props.email}</a>);
-
-  const communityAccessibilitySite = `${props.communityAccessibilityLink}`;
   const communityAccessibilityElement = (
-    <a href={communityAccessibilitySite}>Website Accessibility Policy</a>
+    <a href={props.communityAccessibilityLink}>Website Accessibility Policy</a>
   );
 
   /* eslint-disable max-len */
--- a/src/components/AccessibilityBody/index.jsx
+++ b/src/components/AccessibilityBody/index.jsx
@@ -1,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AccessibilityBody = (props) => {
+  const mailto = `mailto:${props.email}`;
+  const emailElement = (<a href={mailto}>{props.email}</a>);
+
+  const communityAccessibilitySite = `${props.communityAccessibilityLink}`;
+  const communityAccessibilityElement = (
+    <a href={communityAccessibilitySite}>Website Accessibility Policy</a>
+  );
+
+  /* eslint-disable max-len */
+  return (
+    <div>
+      <h1>Studio Accessibility Policy</h1>
+      <p>
+        At edX, we seek to understand and respect the unique needs and perspectives of the edX global community.  We value every course team and are committed to being a leader in expanding access to all, including course team creators and authors with disabilities.  To that end, we have adopted a {communityAccessibilityElement} and this process to ensure that course team creators and authors with disabilities are able to develop and post content on our platform via Studio.
+      </p>
+      <p>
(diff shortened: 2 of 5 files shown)


dsjen:
I just realized that I should probably wrap this in a bootstrap grid. Feel free to review still, and I'll update this in a bit.

src/components/AccessibilityBody/index.jsx
```diff
@@ -0,0 +1,61 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AccessibilityBody = (props) => {
+  const mailto = `mailto:${props.email}`;
```
MichaelRoytman: Should this be `mailTo`? Or is it clearer to leave it as mailto, since it's referring to the URI?

src/components/AccessibilityBody/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+  const emailElement = (<a href={mailto}>{props.email}</a>);
+
+  const communityAccessibilitySite = `${props.communityAccessibilityLink}`;
```
MichaelRoytman: I see that `communityAccessibilityLink` is a String in the PropTypes. Can you explain the use of the template string? Does` const communityAccessibilitySite = props.communityAccessibilityLink;` not work?

src/components/AccessibilityBody/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+  const communityAccessibilityElement = (
+    <a href={communityAccessibilitySite}>Website Accessibility Policy</a>
+  );
+
+  /* eslint-disable max-len */
```
MichaelRoytman: Out of curiosity, why disable the max-len rule? Does breaking the long text into shorter lines affect how it's displayed in the browser?

src/components/AccessibilityPolicyPage/index.jsx
```diff
@@ -3,6 +3,8 @@ import { connect } from 'react-redux';
 import PropTypes from 'prop-types';
 
 // import styles from './AccessibilityPolicyPage.scss';
+import AccessibilityBody from '../AccessibilityBody';
```
MichaelRoytman: I just realized that the form is probably technically part of the "body", but I think that's probably super inconsequential.

src/components/AccessibilityBody/AccessibilityBody.test.jsx
```diff
@@ -0,0 +1,25 @@
...
+        <AccessibilityBody
+          communityAccessibilityLink="http://example.com"
+          phoneNumber="555-555-5555"
+          email="example@example.com"
+        />,
+      );
+    });
+    it('contains links', () => {
+      expect(wrapper.text()).toContain('example@example.com');
+      expect(wrapper.text()).toContain('555-555-5555');
```
MichaelRoytman: Is there a way to add an expect for the communityAccessibilityLink?

dsjen:
FYI, I just pushed changes to wrap the page in a bootstrap container and also added the a11y and assets page urls to the readme.

src/components/AccessibilityBody/index.jsx
dsjen: Yeah, I was sort of debating that, and just went with the lowercase one as it's consistent with the attribute. I'm inclined to leave it as is.

src/components/AccessibilityBody/index.jsx
dsjen: Good catch! I was futzing around with something and can get rid of this. I'll update to:
```
  const communityAccessibilityElement = (
    <a href={props.communityAccessibilityLink}>Website Accessibility Policy</a>
  );
```

src/components/AccessibilityBody/index.jsx
dsjen: It doesn't affect the display, but I didn't want to have to muck around with formatting the text essentially. So...disabling the line count is partially out of laziness, but also in that we get the copy from legal and I'd rather just copy/paste.

src/components/AccessibilityBody/AccessibilityBody.test.jsx
dsjen: Sure thing!

I'll add in something like:
```
      expect(wrapper.find({ href: 'http://example.com' })).toHaveLength(1);
```

MichaelRoytman:
Thanks for adding documentation around accessing this page, too!

src/components/AccessibilityBody/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+const AccessibilityBody = (props) => {
+  const mailto = `mailto:${props.email}`;
+  const emailElement = (<a href={mailto}>{props.email}</a>);
+  const communityAccessibilityElement = (
+    <a href={props.communityAccessibilityLink}>Website Accessibility Policy</a>
+  );
+
+  /* eslint-disable max-len */
+  return (
```
jaebradley: Does all this text need to be internationalized? Or is that not in scope or relevant?

MichaelRoytman: I don't believe it needs to be internationalized. Only the link in the footer to this page needs to be.

src/components/AccessibilityBody/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+          <li>How soon you need access and for how long (e.g., a planned course start date or in connection with a course-related deadline such as a final essay).</li>
+        </ul>
+        <li>Within 48 business hours of receiving your request, the edX Support Team will respond to confirm receipt and forward your request to the edX Partner Manager for your institution and the edX Website Accessibility Coordinator.</li>
+        <li>With guidance from the Website Accessibility Coordinator, edX will contact you within 5 business days to discuss your request and gather additional information from you to identify a solution.</li>
+        <li>EdX will assist you promptly and thoroughly so that you are able to create content on the CMS within your time constraints.  Such efforts may include, but are not limited to: (a) purchasing a third-party tool or software for use on an individual basis to assist your use of Studio, (b) engaging a trained independent contractor to provide real-time visual, verbal and physical assistance, or (c) developing new code to implement a technical fix.</li>
```
jaebradley: Super (Grammar) Nit: it's `edX` in some places and `EdX` in others - I think it should be `edX`?

src/components/AccessibilityBody/index.jsx
jaebradley: You could also just forgo creating the variable and just stick it in the `href` directly. It doesn't look like `mailto` is being used anywhere else.

src/components/AccessibilityBody/index.jsx
```diff
@@ -0,0 +1,59 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AccessibilityBody = (props) => {
+  const mailto = `mailto:${props.email}`;
+  const emailElement = (<a href={mailto}>{props.email}</a>);
```
jaebradley: Just FYI, we now have a [`Hyperlink` `Paragon` component](https://github.com/edx/paragon/blob/master/src/Hyperlink/index.jsx)

dsjen:
@jaebradley -- this PR is in regards to specific legal text.

PR #59, opened 2017-12-01
Updated a11y policy sublists to be alphabetically enumerated.

dsjen:
@cptvitamin requested that both sub-lists be alphabetically enumerated. Here it is!

Please review, @edx/educator-dahlia

### DIFF
--- a/src/components/AccessibilityBody/AccessibilityBody.scss
+++ b/src/components/AccessibilityBody/AccessibilityBody.scss
@@ -1,0 +1,3 @@
+.alphabetical-list {
+  list-style-type: lower-alpha;
+}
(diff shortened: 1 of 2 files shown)


src/components/AccessibilityBody/AccessibilityBody.scss
```diff
@@ -0,0 +1,3 @@
+.alphabetical-list {
+  list-style-type: lower-alpha;
+}
```
fysheets: Awesome! I didn't know this was a thing 👍

dsjen: Mark pointed it out to me!

PR #61, opened 2017-12-04
Accessibility Policy Form

fysheets:
##  [EDUCATOR-1828](https://openedx.atlassian.net/browse/EDUCATOR-1828)


src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <div>
+        <h1 id="a11y-feedback" className={styles['page-header']}>Studio Accessibility Feedback</h1>
+        {this.renderStatusAlert()}
+        <h6>All fileds required.</h6>
```
dsjen: Spelling :)

Also, I don't think `h6` is semantically right for this. Screen readers sometimes navigate by headers and this should probably a div with some styling.

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+    if (!validFormContent) {
+      status = {
+        alertType: 'danger',
+        alertDialog: (
+          <div>
+            <div>Please fill in all required fields.</div>
+            <br />
+            <div>
+              <ul>
```
dsjen: Might need to add some bullet styles to this, similar to https://github.com/edx/studio-frontend/blob/master/src/components/AccessibilityBody/AccessibilityBody.scss#L13.

e.g. `list-style-type: circle;`

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+    this.handleChange = this.handleChange.bind(this);
+    this.validateFormContent = this.validateFormContent.bind(this);
+    this.closeStatusAlert = this.closeStatusAlert.bind(this);
+    this.renderStatusAlert = this.renderStatusAlert.bind(this);
+  }
+
+  onSubmitClick() {
+    this.props.clearAccessibilityStatus();
+    this.setState({ // this will clear on every submit (success/fail) - correct?
```
dsjen: I don't think this is actually clearing out the form. I'm seeing some oddness when clicking submit with an invalid form, changing a field, then clicking submit again.

Also, I'm not sure that we'd want to clear out the form when a submission is invalid as users would have an easier time correcting a incorrectly filled out form.

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+    return status;
+  }
+
+  validateFormContent(inputType, inputValue) {
+    let feedback = { isValid: true };
+    if (inputType === 'email') {
+      /* eslint-disable max-len */
+      /* eslint-disable no-useless-escape */
+      const emailRegEx = /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
```
jaebradley: This regex fails for the given valid email address according to RFC2822
> !#$%&'*+-/=.?^_`{|}~@[IPv6:0123:4567:89AB:CDEF:0123:4567:89AB:CDEF]
> (from [this StackOverflow answer](https://stackoverflow.com/a/297494)).

I know that email address might seem absurd, and this regex is probably good for 99.99999% of all business cases but the point is why not use a library for email validation instead of having to keep (and maintain) this validation regex?

There's [`isemail`](https://www.npmjs.com/package/isemail), which has been downloaded ~4 million times in the last month and has one dependency, `punycode`, which itself has no dependencies.

__`isemail` does validate the previous email address to `true`__
image ([Link](https://runkit.com/embed/4ysueel4j9r8))

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+  onSubmitClick() {
+    this.props.clearAccessibilityStatus();
+    this.setState({ // this will clear on every submit (success/fail) - correct?
+      formEmail: '',
+      formFullName: '',
+      formMessage: '',
+    });
+
+    const { formEmail, formFullName, formMessage } = this.state;
+    const validateEmail = this.validateFormContent('email', formEmail);
```
jaebradley: Why have one validation method? Why not have a `validateEmailEmail` and a `validateName` and a `validateMessage` method? You're essentially doing this by`if...else-if`ing at the top level of `validateFormContent` - why not move that logic into their own functions? The benefits I see are that it makes testing easier as you're testing smaller modules or logic, and it (slightly) increases readability by taking a relatively generic name like `validateFormContent` and making it more specific (`validateEmailAddress`).

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+
+import { clearAccessibilityStatus, submitAccessibilityForm } from '../../data/actions/accessibility';
+import { accessibilityActions } from '../../data/constants/actionTypes';
+import styles from './AccessibilityPolicyForm.scss';
+
+export class AccessibilityPolicyForm extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      formEmail: '',
```
jaebradley: Naming nit: this class has `Form` in the title - might it make sense to remove `form` from the names of the internal variables?

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+      validFormContent: validContent,
+      validationMessages: [
+        validateEmail.validationMessage,
+        validateFullName.validationMessage,
+        validateMessage.validationMessage,
+      ],
+    });
+
+    if (validContent) {
+      this.props.submitAccessibilityForm(formEmail, formFullName, formMessage);
```
jaebradley: Naming nit: this class is called `AccessibilityPolicyForm` so maybe a prop called `submitAccessibilityForm` is slightly redundant - why not `onSubmit`?

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+        <h1 id="a11y-feedback" className={styles['page-header']}>Studio Accessibility Feedback</h1>
+        {this.renderStatusAlert()}
+        <h6>All fileds required.</h6>
+        <section className={styles['form-section']} role="group" aria-labelledby="a11y-feedback">
+          <InputText
+            name="email"
+            label="Your Email"
+            id="email"
+            type="email"
+            onChange={this.handleChange}
```
jaebradley: Same thing here - why not a `handleEmailAddressChange`?

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+        dialog={status.alertDialog}
+        open={this.state.statusAlertOpen}
+        onClose={this.closeStatusAlert}
+        ref={(input) => { this.statusAlertRef = input; }}
+      />
+    );
+
+    return (
+      <div>
+        {statusAlert}
```
jaebradley: Why not skip the variable assignment and just return specify the `StatusAlert` component here? Do you need to wrap it in a `div` or can you just return it?

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+            value={this.state.messageInputRef}
+            themes={['danger']}
+            validator={message => this.validateFormContent('message', message)}
+            inputRef={(ref) => { this.messageInputRef = ref; }}
+          />
+          <Button
+            buttonType="primary"
+            label="Submit"
+            aria-label="Submit Accessibility Feedback Form"
+            onClick={() => { this.onSubmitClick(); }}
```
jaebradley: So correct me if I'm wrong but it looks each of the form areas have their own validators, and then we also do the same validation in `onSubmitClick`, right? Why not block the ability to click this button at all if any of the text areas have invalid inputs?

src/data/actions/accessibility.js
```diff
@@ -0,0 +1,41 @@
...
+      .then((response) => {
+        // if 429 - Rate limit hit
+        // fetch response json to get more details
+        // else show user success message as server handles all other requests properly
+        if (response.status === 429) {
+          return response.json();
+        }
+
+        dispatch(submitAccessibilityFormSuccess(response));
+        return undefined;
```
jaebradley: is this explicit `return` necessary? Doesn't `.then()` return a `Promise` with an `undefined` value if you don't specify a `return` value?

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -0,0 +1,278 @@
...
+      formEmail: '',
+      formFullName: '',
+      formMessage: '',
+    });
+
+    const { formEmail, formFullName, formMessage } = this.state;
+    const validateEmail = this.validateFormContent('email', formEmail);
+    const validateFullName = this.validateFormContent('fullName', formFullName);
+    const validateMessage = this.validateFormContent('message', formMessage);
+    const validContent = validateEmail.isValid &&
```
jaebradley: Naming nit: if this represents a `boolean` why not use a name like `isValidContent` here? Also, using a name like `validFormContent` (below) to represent a `boolean` can make it slightly confusing to read as one could expect `validFormContent` to represent a validated `string`.

src/components/AccessibilityPolicyForm/index.jsx
fysheets: Using a library for this to prevent any need for future maintainability is a good idea -- I think for this MVP we'll stick with the regex we have currently, but I'll add a TODO to investigate using that package at a future version, thanks!

src/components/AccessibilityPolicyForm/index.jsx
fysheets: There are different validation view updates -- the input box changes onBlur to validate (border and validation message below) - but there was also an ask for a view of a Status Alert that will also call out any errors when trying to submit that will alert the users that things are incorrect and to go back and resolve.

src/data/actions/accessibility.js
fysheets: Ah thanks! I wanted to discuss with the team about the proper way to handle this situation. I will come back to this once some other clean up has been done.

src/components/AccessibilityPolicyForm/index.jsx
fysheets: The idea here is if there was going to be additional styling necessary around the status alert to be implemented. I'm going to leave this as is for now, but once we've confirmed the final styles if this is still as is, we can move it to avoid the extra declaration.

src/components/AccessibilityPolicyForm/index.jsx
fysheets: This makes sense and is easy enough to clean up for readability - I am going to make `formEmail` something like `submitterEmail` though because `email` is a prop that is already passed in to handle the a11y email address (I'm going to change that to be `accessibilityEmail` for additional clarity)

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+    it('hides StatusAlert on initial load', () => {
+      const statusAlert = wrapper.find('StatusAlert');
+      expect(statusAlert.find('div').first().prop('hidden')).toEqual(true);
+    });
+
+    it('adds validation checking on each input field', () => {
+      const emailInput = wrapper.find('input#email');
+      emailInput.simulate('blur');
+      // expect(spy).toHaveBeenCalledTimes(1);
```
MichaelRoytman: Do you need this line? If not, you should remove it.

MichaelRoytman: There's a few commented lines in this file.

fysheets: That can be removed, it was mostly to remind me to ask someone on the team how to properly leverage mock functions in tests :) I feel like I've mostly found work arounds for doing so for right now and want to to see if I could be writing these tests in a better way.

MichaelRoytman: Does jest.fn() not work as usual?

fysheets: It probably works as it usually has, but it wasn't working as I was expecting, so I'm guessing that I was making some weird assumptions

MichaelRoytman: If it's something that you'd like to tackle this sprint and would like another eye, I'm happy to take a look, if I can be helpful. If not, can we make a note of it somewhere so we can make sure to resolve it?

PR #66, opened 2017-12-06
Update styles to look better in Studio.

dsjen:
**How it looks in my local Studio**
screen shot 2017-12-06 at 1 56 58 pm

@edx/educator-dahlia

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.5.0",
+  "version": "0.5.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.5.0",
+  "version": "0.5.1",
```
dsjen: I went ahead and bumped the version in this PR. My intention is to make a release, but can certainly pull this into a separate PR there are other ones landing.

PR #69, opened 2017-12-08
Clear A11y Form on Successful Submit

fysheets:
Integrate new Paragon package to handle clearing of form on successful submit

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.6.0",
+  "version": "0.6.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^0.4.0",
-    "@edx/paragon": "^1.4.5",
+    "@edx/paragon": "^1.4.6",
```
efischer19: Let's go ahead and bump SFE to `0.6.1` with this change

PR #70, opened 2017-12-11
fix(email-validation): use email validation library instead of regex

jaebradley:
## Introduction
Related to [this conversation](https://github.com/edx/studio-frontend/pull/61#discussion_r155523898) from #61.

Instead of using regex for email validation, we should use a library ([`isemail`](https://www.npmjs.com/package/isemail), for example).

## Testing Locally
image

image

cc: @edx/educator-dahlia

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "classnames": "^2.2.5",
     "copy-to-clipboard": "^3.0.8",
     "font-awesome": "^4.7.0",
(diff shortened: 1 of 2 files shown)


fysheets:
Closing this as SFE is not being actively maintained. When the repo is updated to match the rest of our system this will be picked up as upgrade work.

PR #71, opened 2017-12-11
Add additional tests to cover Accessibility page

fysheets:
# [EDUCATOR-1926](https://openedx.atlassian.net/browse/EDUCATOR-1926)

Additional Coverage for Accessibility Policy Page Changes

### DIFF
--- a/src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
+++ b/src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
@@ -1,9 +1,13 @@
 import React from 'react';
 import Enzyme from 'enzyme';
-import { AccessibilityPolicyForm } from './index';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+
+import WrappedAccessibilityPolicyForm, { AccessibilityPolicyForm } from './index';
 import { accessibilityActions } from '../../data/constants/actionTypes';
 
 const mount = Enzyme.mount;
+const mockStore = configureStore();
 
 const defaultProps = {
   accessibilityStatus: {},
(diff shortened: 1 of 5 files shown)


fysheets:
This currently adds coverage to 100% of the new reducer and action creator file for the accessibility page. As well as ups the coverage for the AccessibilityPolicyForm to 97.5%. The only two lines not in coverage at this time on the form page are the two lines for the mapDispatchToProps functions. I'm investigating how we can cover those as well.

MichaelRoytman:
Does this way of testing mapDispatchToProps work for your case? https://github.com/edx/studio-frontend/pull/54/files#diff-e5b640571dcdc3f1e4ba88171ef62c38R50

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
```diff
@@ -1,9 +1,13 @@
 import React from 'react';
 import Enzyme from 'enzyme';
-import { AccessibilityPolicyForm } from './index';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+
+import WrappedAccessibilityPolicyForm, { AccessibilityPolicyForm } from './index';
```
MichaelRoytman: So, does this mean we should start using the redux-mock-store library to fake the redux store instead of exporting the unwrapped component from now on? If so, I think we should put a ticket in the backlog or do something to change the other tests to match so that we don't follow the old pattern from other test files.

fysheets:
@MichaelRoytman I can try! The way I did it was a way that I had found when searching around online. Let me give the way we have a shot to see if we can have uniformity across the tests :)

MichaelRoytman:
If the way I did it is not in line with the approach you're taking, I'm happy to follow your approach if you think it's the right way to go!

fysheets:
After working with @MichaelRoytman we added cleaner testing of mapDispatchToProps and have agreed to revisit more complete testing of the async function actions and to revisit in the arch meeting whether we should test the function directly or if we want to test with the wrapped component. Once tests complete I'll bump the version and merge/release these changes.

PR #73, opened 2017-12-11
Accessibility Feedback Updates

fysheets:
# [EDUCATOR-1951](https://openedx.atlassian.net/browse/EDUCATOR-1951)

Including: using h2 instead of h1, adding a final period to error text, and moving a nested ol into an enclosing li correctly

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.6.1",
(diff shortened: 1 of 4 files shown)


fysheets:
FYI @edx/educator-dahlia I'm updating migrations locally to test this within the Studio container to validate stylings. Also should this get any sort of release bump?

PR #75, opened 2017-12-14
Specify danger icon description in a11y form

thallada:


dsjen:
Just in case you're not on it already, we'll need to bump the platform too!

thallada:
Yup, I was planning on doing that in this PR: https://github.com/edx/edx-platform/pull/16847

dsjen:
Ah, thank you!

PR #76, opened 2017-12-18
Label form section w/ all fields required msg

thallada:
Screen-reader (orca) now reads out: "Studio Accessibility Feedback All fields are required region" when initially tabbing into the form.

FYI: @cptvitamin


cptvitamin:
@thallada nice work!  Thanks for taking care of this before I could even create a ticket!

PR #78, opened 2017-12-22
add action creator and reducer tests

MichaelRoytman:


src/data/actions/assets.js
```diff
@@ -41,7 +46,7 @@ export const getAssets = (request, courseDetails) =>
         }
       })
       .catch((error) => {
-        dispatch(assetDeleteFailure(error));
+        dispatch(requestAssetsFailure(error));
```
MichaelRoytman: I have changed this to be a requestAssetsFailure, since it's a failure to request assets and not a failure to delete an asset.

src/data/actions/assets.js
```diff
@@ -70,10 +75,15 @@ export const deleteAsset = (assetId, courseDetails) =>
     clientApi.requestDeleteAsset(courseDetails.id, assetId)
       .then((response) => {
         if (response.ok) {
-          dispatch(deleteAssetSuccess(assetId, response));
-        } else {
-          dispatch(assetDeleteFailure(response));
+          return response.json();
```
MichaelRoytman: I have changed the response to be JSON.

src/data/actions/assets.js
```diff
@@ -95,16 +105,20 @@ export const toggleLockAssetFailure = (asset, response) => ({
 export const toggleLockAsset = (asset, courseDetails) =>
   (dispatch) => {
     dispatch(togglingLockAsset(asset));
-    clientApi.requestToggleLockAsset(courseDetails.id, asset)
+    return clientApi.requestToggleLockAsset(courseDetails.id, asset)
```
MichaelRoytman: We have to return something.

src/data/actions/assets.js
```diff
@@ -95,16 +105,20 @@ export const toggleLockAssetFailure = (asset, response) => ({
...
+      })
+      .catch((error) => {
+        dispatch(toggleLockAssetFailure(asset, error));
       });
   };
 
-export const clearAssetsStatus = () =>
-  dispatch =>
-    dispatch({ type: assetActions.CLEAR_ASSETS_STATUS });
+export const clearAssetsStatus = () => ({
```
MichaelRoytman: I don't think this should be an async action creator, so I have changed it to be synchronous.

src/data/actions/assets.test.jsx
```diff
@@ -24,18 +24,23 @@ const middlewares = [thunk];
...
+  });
+
   it('returns expected state from requestAssetsSuccess', () => {
     const expectedAction = { data: 'response', type: assetActions.REQUEST_ASSETS_SUCCESS };
     expect(store.dispatch(actionCreators.requestAssetsSuccess('response'))).toEqual(expectedAction);
   });
   it('returns expected state from assetDeleteFailure', () => {
     const expectedAction = { response: 'response', type: assetActions.DELETE_ASSET_FAILURE };
-    expect(store.dispatch(actionCreators.assetDeleteFailure('response'))).toEqual(expectedAction);
+    expect(store.dispatch(actionCreators.deleteAssetFailure('response'))).toEqual(expectedAction);
```
MichaelRoytman: I just re-ordered the words to be more consistent with other action creators.

src/data/reducers/assets.js
```diff
@@ -6,31 +6,31 @@ import { assetLoading } from '../constants/loadingTypes';
 import { addLoadingField, removeLoadingField, toggleLockAsset } from './utils';
 import { getAssetAPIAttributeFromDatabaseAttribute } from '../../utils/getAssetsAttributes';
 
-const filtersInitial = {
+export const filtersInitial = {
```
MichaelRoytman: I'm exporting all of the reducers so that I can import and test them.

src/data/reducers/assets.js
```diff
@@ -87,8 +87,18 @@ const sort = (state = sortInitial, action) => {
   }
 };
 
-const status = (state = {}, action) => {
+export const status = (state = {}, action) => {
   switch (action.type) {
+    case assetActions.REQUEST_ASSETS_SUCCESS:
```
MichaelRoytman: I'm updating the status state based on the response of the request.

src/data/actions/accessibility.test.jsx
```diff
@@ -16,19 +16,50 @@ const middlewares = [thunk];
 const mockStore = configureStore(middlewares);
 let store;
 
-describe('Accessibility Action Creator', () => {
+describe('Accessibility Action Creators', () => {
+  beforeEach(() => {
+    store = mockStore(initialState);
```
MichaelRoytman: I'm pulling this out into a beforeEach, since each test case shares the same initial mock store state.

src/data/actions/assets.js
```diff
@@ -15,7 +15,12 @@ export const requestAssetsSuccess = response => ({
   data: response,
 });
 
-export const assetDeleteFailure = response => ({
+export const requestAssetsFailure = response => ({
```
MichaelRoytman: I'm adding a request assets failure action, since we didn't have one before.

src/data/reducers/assets.test.jsx
```diff
@@ -0,0 +1,290 @@
+import * as reducers from './assets';
+import { assetActions } from '../../data/constants/actionTypes';
+import { assetLoading } from '../../data/constants/loadingTypes';
+import { getAssetAPIAttributeFromDatabaseAttribute } from '../../utils/getAssetsAttributes';
+
+let defaultState;
+
+describe('Assets', () => {
+  describe('assets reducer', () => {
```
MichaelRoytman: This refers to the assets state. I know it looks like it's redundant, but it's the assets reducer of the assets.js file.

MichaelRoytman:
@edx/educator-dahlia This is ready for a review! Thanks!

src/data/actions/assets.js
fysheets: Question here -- if the response from the API to delete the asset comes back with a non 2XX response, does it return any additional info as to why it failed to delete in the body? i.e could we still use a `response.json()` to note a more verbose error message to the user in the future? Or do all non-2XX responses return an empty body?

MichaelRoytman: The API only returns 204 or 404, neither of which have bodies, so I have actually removed the call to jsonify the response. Does that seem like the right call?

MichaelRoytman:
@fysheets I believe I have addressed the issue you brought up. I have removed call to jsonify the response for deleting an asset, since the response will be either 204 or 404, neither of which have bodies. It sounds like, from the research I have done, the failure of the json call is intentional, so we should just avoid calling json on responses without bodies.

fysheets:
@MichaelRoytman got it -- and sure, I'll check it out again thanks for the investigation!

PR #79, opened 2017-12-22
Added dropzone and file upload

dsjen:
@edx/educator-dahlia -- This adds a drag and drop area to the files and uploads page. Here are some changes at a high level:

* added drag and drop that is styled with input from UX, accepts multiple file uploads, includes a button to open the native file selector, limits file uploads by individual file sizes, and total file count
* status is updated when loading files and while file upload is complete along with a table refresh
* refactored the status display logic slightly by removing the state indicating whether it was closed/opened as we can use the presence or absence of the status state to indicate whether something should be shown.

~~I still need to write tests and please feel free to review early if you'd like.~~ Please review!


dsjen:
screen shot 2017-12-22 at 11 02 33 am

public/index.html
```diff
@@ -21,6 +21,7 @@
     id: "course-v1:edX+DemoX+Demo_Course",
     revision: "",
     base_url: "http://localhost:18010",
+    max_file_size_in_mbs: 10,
   },
```
fysheets: This means the release of this will be dependent on the need to pass this in from Studio in the studioContext, right?

dsjen: Yep! I'll need to make a PR to the platform to pass that in. It is [max_file_size_in_mbs](https://github.com/edx/edx-platform/blob/d0687e97b18a54d84c2534256f193a9ace35ab88/cms/templates/asset_index.html#L111).

src/data/actions/assets.js
```diff
@@ -108,3 +108,52 @@ export const toggleLockAsset = (asset, courseDetails) =>
...
+});
+
+export const uploadAssets = (files, courseDetails) =>
+  (dispatch, getState) => {
+    dispatch(uploadingAssets(files.length));
+    files.forEach(file => {
+      clientApi.postUploadAsset(courseDetails.id, file)
+        .then((response) => {
+          if (response.ok) {
+            response.json().then(data => {
```
fysheets: I really like the `.then` here instead of how we've done it in a few other places where it's on the same level as the original `.then` this kind of makes more sense to me when reading it the order of operations of what's happening

src/components/AssetsTable/index.jsx
```diff
@@ -301,7 +304,6 @@ export class AssetsTable extends React.Component {
       deletedAsset,
       elementToFocusOnModalClose: this.statusAlertRef,
       modalOpen: false,
-      statusAlertOpen: true,
     });
```
fysheets: Looks like statusAlertOpen in the state was removed in favor of the ternary on line 414 -- out of curiosity can you explain the logic that went through that decision?

public/index.html
fysheets: One more question as I look at this -- does this make sense to be in courseDetails? Mostly because it's not something that is course specific, or should we have another object for it?

dsjen: Right on! What about `upload_settings`?

src/components/AssetsTable/index.jsx
dsjen: I didn't think it was necessary as the state reflects whether the status alert is rendered. Does that seem like the right approach?

public/index.html
dsjen: I updated this setting to be nested inside of `upload_settings`.

public/index.html
```diff
@@ -22,6 +22,9 @@
     revision: "",
     base_url: "http://localhost:18010",
   },
+  upload_settings: {
+    max_file_size_in_mbs: 10,
+  }
```
fysheets: 👍

src/components/AssetsTable/index.jsx
thallada: I like the simplicity, but we'll need to check if dynamically adding an element with the alert role is accessible or if it needs to be on the page at DOM load to be registered by screen readers.

fysheets: Good catch @thallada

dsjen: It sounds like we actually *do* need to keep the alert element in the page, right? I can update the rendering for this case.

src/components/AssetsTable/index.jsx
```diff
@@ -197,6 +196,38 @@ export class AssetsTable extends React.Component {
...
+    switch (assetsStatus.type) {
+      case assetActions.UPLOAD_ASSET_SUCCESS:
+        return {
+          alertType: 'success',
+          alertDialog: `${assetsStatus.loadedCount} files successfully uploaded.`,
+        };
+      case assetActions.UPLOADING_ASSETS:
+        return {
+          alertType: 'info',
+          alertDialog: `${assetsStatus.count} files uploading.`,
```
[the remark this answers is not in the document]
dsjen: I like this idea! I'll double check w/ docs on this.

src/components/AssetsTable/index.jsx
```diff
@@ -197,6 +196,38 @@ export class AssetsTable extends React.Component {
     };
   }
 
+  getUploadStatusFields(assetsStatus) {
+    switch (assetsStatus.type) {
+      case assetActions.UPLOAD_ASSET_SUCCESS:
+        return {
+          alertType: 'success',
+          alertDialog: `${assetsStatus.loadedCount} files successfully uploaded.`,
```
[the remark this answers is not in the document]
dsjen: So, I like the idea of having a total count. I'm just not sure how to handle the edge cases around it though, e.g.
* Some number of files were uploaded, then another set of files were uploaded. Maybe we can simply keep an aggregated count of files uploaded?
* If the status alert/message is closed, the state is in turn cleared out. We'd need to include logic to persist the various counts.

Maybe what we need is better text? I'll check in w/ docs and report back.

src/components/AssetsTable/index.jsx
thallada: Yep, I checked with Mark and he said that, while some browsers and screen readers are okay with dynamically adding it, we should keep the alert element on the page for compatibility with most screen readers.

src/data/actions/assets.js
```diff
@@ -108,3 +108,52 @@ export const toggleLockAsset = (asset, courseDetails) =>
...
+export const uploadAssets = (files, courseDetails) =>
+  (dispatch, getState) => {
+    dispatch(uploadingAssets(files.length));
+    files.forEach((file) => {
+      clientApi.postUploadAsset(courseDetails.id, file)
+        .then((response) => {
+          if (response.ok) {
+            response.json().then((data) => {
+              dispatch(uploadAssetSuccess(data.asset, response));
+              dispatch(getAssets(getState().request, courseDetails));
```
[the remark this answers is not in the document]
dsjen: I totally "get" this case! I had thought about using skeleton states, but the issues with this is determining whether the file has been filtered by file type, winnowed down by search, paginated, and sorted. It's difficult to tell where the file should go and if it should even be displayed. I like the idea though and think it's feasible--just would take more effort than I'm wanting to put in at the moment. 😞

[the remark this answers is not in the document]
thallada: I had forgotten about the filters and pagination. For what it's worth, I think it would be fine if the uploaded files were inserted at the top of the list until the GET request came back with their actual positions. But yeah, probably not worth the effort for now.

src/components/AssetsTable/index.jsx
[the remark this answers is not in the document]
dsjen: @srpearce -- I used your copy in the files & uploads status message when files are uploading (see image attached). What do you think about adding ellipsis at the end of the message?

screen shot 2018-01-04 at 10 33 20 am

src/components/AssetsTable/index.jsx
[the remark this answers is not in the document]
dsjen: @srpearce -- I've updated the status message when files are successfully uploaded. This happens incrementally as each file is uploaded. I've attached a screenshot of the message.

The issue is that we don't keep track of the total number of files being uploaded, so it might be that 2 out of the 3 files that you've uploaded have completed and you see something like the attached screenshot. In lieu of having a total count of files being uploaded, is there any creative copy updates that we could do to indicate that files have been updated and that there may or may not be more in progress? Is that asking too much?!?

screen shot 2018-01-04 at 10 39 30 am

dsjen:
I've updated this to keep the alert in the DOM (thanks for catching that!) and have rebased and resolved conflicts. I'm planning on merging this and making any copy updates as separate PRs. We'll finally be able to upload! 🙃

src/components/AssetsTable/index.jsx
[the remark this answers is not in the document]
srpearce: @dsjen I can't think of a good way to write around this, unfortunately. The only options I can see are:
* Don't update incrementally—only add a message when the file uploader is finished with everything (definitely not ideal)
* Change the period after "n files successfully uploaded" to an ellipsis if any other files are still uploading, and back to a period when all files are done (also not ideal because it's probably way too subtle for anyone to notice)

How hard would it be to keep track of an aggregated number of files and go with Tyler's suggestion ("n of x files uploaded")? I think that's how most UIs I've seen do it.

src/components/AssetsTable/index.jsx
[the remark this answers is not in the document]
srpearce: I like the idea in theory—the only problem is i18n. Some languages may not have a comparable punctuation mark. Because of that, I think it would be more effective to leave the period and add a spinner.

PR #82, opened 2017-12-29
Update UI to leverage Paragon Warning Variant Modal

fysheets:
[EDUCATOR-2015](https://openedx.atlassian.net/browse/EDUCATOR-2015)


fysheets:
@edx/educator-dahlia Ready for review! :) I think the idea is to do a minor version bump with this change as well.

PR #83, opened 2018-01-05
AssetsFilters Styles

fysheets:
# [EDUCATOR-2048](https://openedx.atlassian.net/browse/EDUCATOR-2048)

### DIFF
--- a/src/components/AssetsFilters/AssetsFilters.scss
+++ b/src/components/AssetsFilters/AssetsFilters.scss
@@ -1,4 +1,9 @@
-ul.filter-set {
-  list-style: none;
-  padding-left: $spacer;
+@import "~bootstrap/scss/utilities/_spacing.scss";
+
+.filter-heading {
+  @extend .pt-3;
+}
+
+.filter-set {
+  padding-left: $spacer * 2;
 }
--- a/src/components/AssetsFilters/AssetsFilters.test.jsx
+++ b/src/components/AssetsFilters/AssetsFilters.test.jsx
@@ -21,10 +21,23 @@ describe('<AssetsFilters />', () => {
       );
     });
     it('correct number of filters', () => {
-      expect(wrapper.find('li')).toHaveLength(5);
+      const checkBoxGroup = wrapper.find('CheckBoxGroup');
+
+      expect(checkBoxGroup).toHaveLength(1);
+      expect(checkBoxGroup.find('[type="checkbox"]')).toHaveLength(5);
     });
     it('correct styling', () => {
-      expect(wrapper.find('ul').hasClass('filter-set')).toEqual(true);
+      expect(wrapper.find('h4')).toHaveLength(1);
+      expect(wrapper.find('h4').hasClass('filter-heading')).toEqual(true);
+      expect(wrapper.find('div').at(1).hasClass('filter-set')).toEqual(true);
+    });
+    it('handles onChange callback correctly', () => {
+      const checkBoxGroup = wrapper.find('CheckBoxGroup');
+      let checkBoxes = checkBoxGroup.find('[type="checkbox"]');
+
+      checkBoxes.first().simulate('change', { target: { checked: true, type: 'checkbox' } });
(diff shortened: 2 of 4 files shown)


fysheets:
@edx/educator-dahlia this should be ready for review :)

src/components/AssetsFilters/AssetsFilters.scss
```diff
@@ -1,4 +1,9 @@
-ul.filter-set {
-  list-style: none;
-  padding-left: $spacer;
+@import "~bootstrap/scss/utilities/_spacing.scss";
+
+.filter-heading {
+	@extend .pt-3;
```
MichaelRoytman: This indentation looks a little funny relative to the indentation below. Should this be two spaces instead of what looks like a tab?

src/components/AssetsFilters/index.jsx
```diff
@@ -30,18 +31,23 @@ const ASSET_TYPES = [
...
-  </ul>
+  <div role="group" aria-labelledby="filter-label">
+    <h4 id="filter-label" className={styles['filter-heading']}>Filter by File Type</h4>
+    <div className={styles['filter-set']}>
+      <CheckBoxGroup>
+        {ASSET_TYPES.map(type => (
+          <CheckBox
+            key={type.key}
+            id={type.key}
+            name={type.key}
```
MichaelRoytman: I didn't follow the discussion around these changes and the CheckBoxGroup while they were happening, so please correct me if I'm wrong. When we were at the Accessibility Unmeeting before the winter holidays, Mark said that the checkbox's ids should all be unique (and they were all the same at the time), and that the names should all be the same (my notes say something like "filters"). Is that correct? Does the CheckBoxGroup solve that problem? I'm wondering whether the name attribute still being unique to the checkbox is a problem?

src/components/AssetsFilters/AssetsFilters.test.jsx
```diff
@@ -21,10 +21,23 @@ describe('<AssetsFilters />', () => {
       );
     });
     it('correct number of filters', () => {
-      expect(wrapper.find('li')).toHaveLength(5);
+      const checkBoxGroup = wrapper.find('CheckBoxGroup');
```
MichaelRoytman: Thanks for updating the tests! 😄

src/components/AssetsFilters/AssetsFilters.scss
MichaelRoytman: I just want to double check: the .pt-3 classname comes from the Bootstrap spacing utilities, right?

fysheets: Good eye :)

fysheets: Yep - https://getbootstrap.com/docs/4.0/utilities/spacing/ I found it from the PR Dennis had in Paragon with table headings.

src/components/AssetsFilters/index.jsx
fysheets: Changing this so that all of the names are the same does not affect functionality - let me sanity check on having those be the same is necessary for Accessibility and get back to you with the final answer here with regards to the `name` attribute :) I do remember that the `id` needed to be unique for sure.

fysheets: The examples here https://www.w3.org/WAI/tutorials/forms/grouping/ show that the name attribute is the same for radio buttons, but not for checkbox inputs. I think we're okay to keep this unique, but will check in with @cptvitamin for final verification. 

screen shot 2018-01-09 at 11 09 29 am

fysheets: Confirmed -- leaving this as is for now. Looking at the latest spec https://www.w3.org/TR/html52/sec-forms.html#checkbox-state-typecheckbox - the name attribute is needed for Radio buttons, but isn't required to be identical for checkboxes.

MichaelRoytman: That sounds good. Thanks a lot for following up on that!

PR #84, opened 2018-01-05
i18n setup

efischer19:
# [EDUCATOR-1611](https://openedx.atlassian.net/browse/EDUCATOR-1611)

Begins the process of extracting display strings everywhere they're defined. the a11y page has been tested and seems to be working; the files and uploads page is in progress.

I think my branch is outdated, I'll rebase and force push.


efischer19:
Screenshot of https://github.com/edx/studio-frontend/pull/84/commits/f4269496eb5ec42be48cb818ed87a07ce923a3df, when I manually "translated" all strings by prepending them with `FRENCH:`

screen shot 2018-01-05 at 1 58 53 pm

efischer19:
I *think* I've gotten them all here. Next steps before review:

- [ ] manually go through SFE source to ensure I got all strings (I think some were added after I cut my branch?)
- [ ] manually evaluate display string names/descriptions
- [ ] enumerate outstanding TODO questions
- [ ] verify that everything still looks the same locally

package.json
```diff
@@ -15,6 +17,8 @@
     "@edx/edx-bootstrap": "^0.4.3",
     "@edx/paragon": "^1.7.1",
     "babel-polyfill": "^6.26.0",
+    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-stage-0": "^6.24.1",
```
efischer19: We're pretty sure there's a better way to do this, per @MichaelRoytman

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -256,21 +269,25 @@ export class AccessibilityPolicyForm extends React.Component {
...
+            { displayText =>
+              (<Button
+                buttonType="primary"
+                label={<WrappedMessage message={messages.common.label_submit} />}
+                aria-label={displayText}
+                onClick={() => { this.onSubmitClick(); }}
+                inputRef={(ref) => { this.submitButtonRef = ref; }}
+              />)
+            }
+          </WrappedMessage>
```
efischer19: Here's how this looks once run, note that there are 2 different `lang` spans, these could be set to different languages (say, if we had translations available for one but not the other)

screen shot 2018-01-29 at 11 09 31 am

src/components/AccessibilityPolicyForm/displayMessages.jsx
```diff
@@ -0,0 +1,86 @@
...
+  },
+  thanks: {
+    id: 'success.thanks',
+    description: 'a simple thank you message',
+    defaultMessage: 'Thank you for contacting edX!',
+  },
+  detailed_thanks: {
+    id: 'success.detailed_thanks',
+    description: 'a longer thank you message with more details',
+    defaultMessage: 'Thank you for your feedback regarding the accessibility of Studio. We typically respond within one business day (Monday to Friday, 13:00 to 21:00 UTC).',
```
efischer19: These dates will need special handling

src/components/AssetsFilters/index.jsx
```diff
@@ -4,43 +4,50 @@ import CheckBoxGroup from '@edx/paragon/src/CheckBoxGroup';
 import CheckBox from '@edx/paragon/src/CheckBox';
 import { connect } from 'react-redux';
 
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
 import { filterUpdate } from '../../data/actions/assets';
 import styles from './AssetsFilters.scss';
+import messages from './displayMessages';
```
efischer19: What, if any, is the standard we use for ordering imports in jsx? It seems like we largely have third party and first party separate, do we go by <thing being imported> or <location we're importing from> after that?

efischer19: Not terribly important, just something I've been wondering

src/components/AssetsTable/displayMessages.jsx
```diff
@@ -0,0 +1,163 @@
...
+  },
+  upload_success: {
+    id: 'upload.success',
+    description: 'message shown after upload finishes',
+    defaultMessage: '{uploaded_count} files successfully uploaded.',
+  },
+  upload_in_progress: {
+    id: 'upload.progress',
+    description: 'status message shown after beginning file upload',
+    defaultMessage: '{uploading_count} files uploading.',
```
efischer19: number, I'll see what we can do here

src/components/AssetsFilters/index.jsx
fysheets: @arizzitano might be able to speak more on what to do with our imports - I know there at least are _some_ eslint rules at play, but I'm not sure if there is a standard that we should be following?

src/components/AccessibilityPolicyForm/index.jsx
```diff
@@ -106,23 +114,35 @@ export class AccessibilityPolicyForm extends React.Component {
...
+            values={{
+              emailLink: <a href={`mailto:${accessibilityEmail}`}>{accessibilityEmail}</a>,
+            }}
+          />
         ),
       };
     } else if (accessibilityStatus.type ===
         accessibilityActions.ACCESSIBILITY_FORM_SUBMIT_SUCCESS) {
+      const start = new Date('Mon Jan 29 2018 13:00:00 GMT (UTC)');
+      const end = new Date('Fri Feb 2 2018 21:00:00 GMT (UTC)');
```
fysheets: Is there any way to make these more dynamic?

efischer19: Yeah, I felt bad about how this ended up. What's the best way to create a `Date` that represents "a Monday" or "1300 UTC on any particular day"?

src/utils/i18n/formattedMessageWrapper.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { FormattedMessage, injectIntl, intlShape } from 'react-intl';
+
+const WrappedMessage = (props) => {
+  const { message, intl, ...other } = props;
+  let locale = intl.defaultLocale;
+  if (message.id in intl.messages) {
+    locale = intl.locale;
+  }
```
efischer19: make a comment here

src/components/AssetsTable/index.jsx
```diff
@@ -370,13 +540,18 @@ export class AssetsTable extends React.Component {
...
-        title={`Delete ${this.state.assetToDelete.display_name}`}
+        title={(<WrappedMessage
+          message={messages.assetsTableDeleteObject}
+          values={{
+            displayName: this.state.assetToDelete.display_name,
+          }}
+        />)}
         body={this.renderModalBody()}
-        closeText="Cancel"
+        closeText={<PlainI18nString message={messages.assetsTableCancel} />}
```
efischer19: This cannot be a `FormattedMessage`, see https://github.com/edx/paragon/pull/138

src/components/AssetsTable/index.jsx
```diff
@@ -368,41 +537,87 @@ export class AssetsTable extends React.Component {
...
+            <Modal
+              open={this.state.modalOpen}
+              title={(<WrappedMessage
+                message={messages.assetsTableDeleteObject}
+                values={{
+                  displayName: this.state.assetToDelete.display_name,
+                }}
+              />)}
+              body={this.renderModalBody()}
+              closeText={displayText}
```
efischer19: This is where I landed on the `closeText` issue. The entire modal is enclosed in a containing `<span lang='en'>`, whose language code will match the rendered language of the "Cancel" string. There are other messages included in the Modal (which could potentially be in a different language); they'll have their own containing spans. The cancel button body and `×` aria-label are the only text that relay on the parent Modal's span:

screen shot 2018-02-01 at 10 27 36 am
screen shot 2018-02-01 at 10 34 49 am

efischer19:
As far as I can tell, the codecov failures are related to my expanding the number of lines contained in already uncovered methods:

- `getLockFailedStatusFields`
- `getDeleteStatusFields`
- `getUploadStatusFields`

[source](https://codecov.io/gh/edx/studio-frontend/src/455831d35efde87cd2a4f96361b89d84d29c5e3f/src/components/AssetsTable/index.jsx)

I'll add a few tests to hit those

efischer19:
ohhhhhh nvm, I had a weird rebase and missed https://github.com/edx/studio-frontend/commit/eb76c98351eaad566e80f9730de949f49e356753#diff-35e343ee67d70894218ba932e51bd1cd. One minute...

efischer19:
@edx/educator-dahlia after some hackathoning, this is finally ready for review. No rush, we can discuss at standup in the morning.

src/utils/i18n/enzymeHelper.jsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * copied from https://github.com/yahoo/react-intl/wiki/Testing-with-React-Intl#enzyme
+ */
+
+import React from 'react';
+import { IntlProvider, intlShape } from 'react-intl';
+import { mount, shallow } from 'enzyme';
+
+// Create the IntlProvider to retrieve context for wrapping around.
+const intlProvider = new IntlProvider({ locale: 'en', messages: {} }, {});
```
MichaelRoytman: I'm trying to understand what `messages` is here, and I can't really find it in the react-intl documentation. What does it do? It almost sounds like it is supposed to use the object in src/data/i18n/locales/{local}.json? Is that right? How does the tested component pull out the correct English string for testing if messages is empty?

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
```diff
@@ -157,7 +156,7 @@ describe('<AccessibilityPolicyForm />', () => {
       expect(wrapper.state('isStatusAlertOpen')).toEqual(true);
       expect(statusAlertType).toEqual('success');
       expect(statusAlert.find('div').first().hasClass('alert-success')).toEqual(true);
-      expect(statusAlert.text()).toContain('Thank you for contacting edX!Thank you for your feedback regarding the accessibility of Studio. We typically respond within one business day (Monday to Friday, 13:00 to 21:00 UTC).');
+      expect(statusAlert.find('div').first().text()).toContain('×Thank you for contacting edX!Thank you for your feedback regarding the accessibility of Studio. We typically respond within one business day (Monday to Friday, 1:00 PM UTC to 9:00 PM UTC).');
```
MichaelRoytman: What is the funny looking x for? Is it the x button of the StatusAlert? It seems kind of odd that it comes through in the call to text().

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
```diff
@@ -28,7 +27,7 @@ const validationMessages = {
   message: 'Enter a message.',
 };
 
-const dangerIconDescription = 'Error: ';
```
MichaelRoytman: I noticed that it seems like a lot of trailing punctuation has been removed from the tests. Why is that? Is there something about internationalizing punctuation?

src/components/AccessibilityPolicyForm/displayMessages.jsx
```diff
@@ -0,0 +1,86 @@
...
+  },
+  accessibilityPolicyFormSuccessDetails: {
+    id: 'accessibilityPolicyFormSuccessDetails',
+    defaultMessage: 'Thank you for your feedback regarding the accessibility of Studio. We typically respond within one business day ({day_start} to {day_end}, {time_start} to {time_end}).',
+    description: 'Detailed thank you message when form submission is successful',
+  },
+  accessibilityPolicyFormValidEmail: {
+    id: 'accessibilityPolicyFormValidEmail',
+    defaultMessage: 'Enter a valid email address.',
+    description: 'Error message an invalid email is entered into the form',
```
MichaelRoytman: I think this could a be a little clearer. Maybe something like `Error message for when an invalid email is entered into the form`?

src/components/AssetsDropZone/displayMessages.jsx
```diff
@@ -0,0 +1,21 @@
...
+    description: 'Upload label for the "Browse" button',
+  },
+  assetsDropZoneHeader: {
+    id: 'assetsDropZoneHeader',
+    defaultMessage: 'Drag and Drop',
+    description: 'Upload with dropzone section heading',
+  },
+  assetsDropZoneMaxFileSizeLabel: {
+    id: 'assetsDropZoneMaxFileSizeLabel',
+    defaultMessage: 'Maximum file size: {maxFileSizeMB} MB',
```
MichaelRoytman: Do we want MB to be translated? Are file sizes different across languages? Will it be translated correctly?

src/components/AssetsTable/AssetsTable.test.jsx
```diff
@@ -566,13 +563,13 @@ describe('Lock asset', () => {
     });
     const statusAlert = wrapper.find('StatusAlert');
     expect(statusAlert.prop('alertType')).toEqual('danger');
-    expect(statusAlert.find('.alert-dialog').text()).toEqual('Failed to toggle lock for marmoset.png.');
+    expect(statusAlert.find('.alert-dialog').text()).toEqual('Failed to toggle lock for marmoset.png');
```
MichaelRoytman: Just flagging an example of the missing punctuation here, which I mentioned above.

src/components/AssetsTable/displayMessages.jsx
```diff
@@ -0,0 +1,161 @@
...
+  },
+  assetsTableNoPreview: {
+    id: 'assetsTableNoPreview',
+    defaultMessage: 'Preview not available',
+    description: 'Label shown when no preview is available',
+  },
+  assetsTableLockedObject: {
+    id: 'assetsTableLockedObject',
+    defaultMessage: 'Locked {thing}',
+    description: 'States that an object has just been locked',
```
MichaelRoytman: Should we be more clear about what "locked" means? I could see it not having the same meaning across languages.

src/components/AssetsTable/displayMessages.jsx
```diff
@@ -0,0 +1,161 @@
...
+  },
+  assetsTableStudioLink: {
+    id: 'assetsTableStudioLink',
+    defaultMessage: 'Studio',
+    description: 'Label that indicates a relative (as opposed to an absolute) URL',
+  },
+  assetsTableWebLink: {
+    id: 'assetsTableWebLink',
+    defaultMessage: 'Web',
+    description: 'Label that indicates a absolute (as opposed to a relative) URL',
```
MichaelRoytman: NIT: a absolute -> an absolute

src/components/AssetsTable/index.jsx
```diff
@@ -132,28 +135,42 @@ export class AssetsTable extends React.Component {
...
     }
-    return (<Button
-      className={classes}
-      label={''}
-      data-asset-id={asset.id}
-      aria-label={`${lockState} ${asset.display_name}`}
-      onClick={this.onLockClick}
-    />);
+    return (
+      <WrappedMessage message={lockStateMessage} values={{ thing: asset.display_name }}>
```
MichaelRoytman: NIT: Maybe something a little clearer than "thing"? "asset"?

src/components/Pagination/Pagination.test.jsx
```diff
@@ -120,12 +119,12 @@ describe('<Pagination />', () => {
     });
     it('is disabled when page is first page', () => {
       expect(previousLink).toHaveLength(1);
-      expect(previousLink.text()).toEqual('previous button disabled');
+      expect(previousLink.text()).toEqual('previousbutton is disabled');
```
MichaelRoytman: Why no more spaces?

src/data/i18n/locales/whitelist_fr.json
```diff
@@ -0,0 +1,2 @@
+[
+]
```
MichaelRoytman: What's the purpose of this file?

codecov.yml
```diff
@@ -0,0 +1,4 @@
+ignore:
+  - "src/accessibilityIndex.jsx"
```
MichaelRoytman: Should we even collect coverage for these files? If not, we could add them to `coveragePathIgnorePatterns` in package.json, and then coverage won't be reported when tests are run locally as well as on codecov.

src/accessibilityIndex.jsx
```diff
@@ -2,20 +2,30 @@ import 'babel-polyfill'; // general ES2015 polyfill (e.g. promise)
...
+import { addLocaleData, IntlProvider } from 'react-intl';
+import enLocaleData from 'react-intl/locale-data/en';
 
 import AccessibilityPolicyPage from './components/AccessibilityPolicyPage';
 import store from './data/store';
 
+addLocaleData(enLocaleData);
+
+// hard-coded NON-defaults for testing. TODO: make these dynamic and correct
+const locale = 'en';
```
MichaelRoytman: Just checking - this is a "do this later" TODO and not a "do this as part of this PR" TODO, right?

src/components/Pagination/index.jsx
```diff
@@ -20,27 +21,29 @@ export class Pagination extends React.Component {
...
+        { displayText => <span className={paginationStyles['sr-only']}>{displayText}</span> }
+      </WrappedMessage>
     );
   }
 
   getPreviousLabel(totalPages) {
     return (
-      <span>
-      previous
+      <React.Fragment>
```
MichaelRoytman: Nice!

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
fysheets: This specific one I can answer -- This text was for the screenreader only text, which doesn't need punctuation (doesn't get read anyway) also trailing punctuation doesn't always work the same way in each language / might not make sense at all with how that language uses punctuation at all. This one specifically was mostly for general clean up purposes, but we can revisit other instances as well.

src/components/AssetsTable/AssetsTable.test.jsx
fysheets: This one should probably be reverted back to the `.`

codecov.yml
efischer19: Cool, TIL. Thanks!

package.json
```diff
@@ -13,8 +15,10 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^0.4.3",
-    "@edx/paragon": "^2.0.0",
+    "@edx/paragon": "^2.0.1",
     "babel-polyfill": "^6.26.0",
+    "babel-preset-es2015": "^6.24.1",
+    "babel-preset-stage-0": "^6.24.1",
```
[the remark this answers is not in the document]
efischer19: I have no idea, TBH. `babel-preset-env` took care of things though, so this is now removed

src/accessibilityIndex.jsx
efischer19: Correct, this will be part of the "get it working with platform" work in EDUCATOR-1613

src/components/AccessibilityBody/displayMessages.jsx
```diff
@@ -0,0 +1,106 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  a11yBodyPolicyLink: {
```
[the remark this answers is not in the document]
efischer19: We've attempted to "prefix" them already, insofar as all the messages for a given page have the same root (`a11yBody` in this case). There was an issue with the tooling not allowing nested dicts in `defineMessages` though, so they're all flat.

Is there some other form of prefixing we should add at this point?

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
efischer19: It's this `&times;` entity: https://github.com/edx/paragon/blob/b783247b0133fc5c64d2d58561c08541cb3bf0b3/src/StatusAlert/index.jsx#L75

`statusAlert` is this whole div, so the `x` icon is included in the return value of text(): 
screen shot 2018-02-05 at 12 22 34 pm

I figured that it would make for a more complete test to include this value in the check, we can easily remove if if we want though since this is a `toContain` check

src/components/AssetsDropZone/displayMessages.jsx
efischer19: I think this will be okay, since the wiki page mentions it being international multiple times in the lead paragraph: https://en.wikipedia.org/wiki/Megabyte

I'll follow up with Bill though at some point, that is a good question

src/components/AssetsTable/displayMessages.jsx
efischer19: I'm good with "the docs define this term, we assume you're read them and agree with us on that definition", but we can clarify more if we want.

src/components/Pagination/Pagination.test.jsx
efischer19: This string previously had an odd space in it: https://github.com/edx/studio-frontend/pull/84/files/#diff-19113b55aba41492c4d2adb0959b1478L23

The full "sentence" isn't really a sentence at all, it's the `previous` label with some follow-up sr-only text, so the text() return value is smashed together like this. 
screen shot 2018-02-05 at 12 35 36 pm

src/data/i18n/locales/whitelist_fr.json
efischer19: output from the `i18n_manage` tooling. It can be used to mark strings that may have a valid translation in transifex, but we want to keep the default untranslated for some reason (a name perhaps?)

src/utils/i18n/enzymeHelper.jsx
efischer19: `messages` here should contain a mapping of {`id`: "`translated string value`"} pairs to use in translation. If a given id is not found in the messages dict (say, if messages is empty), react-intl will fallback to the defined `defaultMessage` value, which is always english in our case/when testing.

src/components/AccessibilityBody/displayMessages.jsx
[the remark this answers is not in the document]
MichaelRoytman: I think for some reason I thought we were going to use periods as delimiters, so I was like, "where are the periods?!" but I probably made that up, so disregard my question :P.

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
MichaelRoytman: Ah, I see. Thanks for the clarification!

src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.test.jsx
MichaelRoytman: Ah, I see. That makes sense. I just didn't realize that the div would include the x in the button. I thought it was just testing the alert-dialog.

src/utils/i18n/enzymeHelper.jsx
MichaelRoytman: Okay, great. Thanks! That's helpful.

codecov.yml
MichaelRoytman: Nit: I'm not sure if we need this file anymore because you added it to package.json, but if you don't feel like rebasing again, it's probably fine to leave.

dsjen:
Hooray! 🦁

PR #86, opened 2018-01-08
Updated table styles

dsjen:
* styled no image preview styles
* added space around copy icon
* styled the buttons

screen shot 2018-01-08 at 5 44 38 pm

@edx/educator-dahlia

### DIFF
--- a/src/components/AssetsTable/index.jsx
+++ b/src/components/AssetsTable/index.jsx
@@ -9,6 +9,7 @@ import classNames from 'classnames';
 import { connect } from 'react-redux';
 
 import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
+import styles from './AssetsTable.scss';
 import { assetActions } from '../../data/constants/actionTypes';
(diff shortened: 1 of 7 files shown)


src/components/AssetsTable/index.jsx
```diff
@@ -159,9 +164,11 @@ export class AssetsTable extends React.Component {
   }
 
   getLoadingLockButton(asset) {
-    const classes = [FontAwesomeStyles.fa, FontAwesomeStyles['fa-spinner'], FontAwesomeStyles['fa-spin']];
+    // spinner classes are applied to the span to keep the whole button from spinning
```
thallada: Haha, that would be pretty fun though

thallada:
Looks like there are some eslint and weird jest failures though

src/components/AssetsTable/index.jsx
dsjen: Ha! I did find it quite amusing in fact! 😸

PR #88, opened 2018-01-09
add additional actions/assets.js, reducers/assets.js, and reducers/utils.js tests

MichaelRoytman:


src/data/actions/assets.js
```diff
@@ -73,8 +73,8 @@ export const deleteAssetFailure = assetId => ({
 export const deleteAsset = (assetId, courseDetails) =>
   dispatch =>
     clientApi.requestDeleteAsset(courseDetails.id, assetId)
-    // since the API returns 204 on success and 404 on failure, neither of which have
-    // content, we don't json-ify the response
+      // since the API returns 204 on success and 404 on failure, neither of which have
```
MichaelRoytman: Just indenting these comments correctly - nothing major.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
           }
-        });
-    });
+          dispatch(uploadAssetFailure(file, response.status));
+          return undefined;
+        })
+    )));
   };
 
-export const uploadExceedMaxCount = maxFileCount =>
```
MichaelRoytman: I don't think that this action creator needs to be asynchronous, so I've changed it to be synchronous.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
-    dispatch({
-      type: assetActions.UPLOAD_EXCEED_MAX_COUNT_ERROR,
-      maxFileCount,
-    });
+export const uploadExceedMaxCount = maxFileCount => ({
+  type: assetActions.UPLOAD_EXCEED_MAX_COUNT_ERROR,
+  maxFileCount,
+});
 
-export const uploadExceedMaxSize = maxFileSizeMB =>
```
MichaelRoytman: I don't think that this action creator needs to be asynchronous, so I've changed it to be synchronous.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
+  asset,
   response,
 });
 
 export const uploadAssets = (files, courseDetails) =>
   (dispatch, getState) => {
     dispatch(uploadingAssets(files.length));
-    files.forEach((file) => {
+    // gather all the promises into a single promise that can be returned
+    return Promise.all(files.map(file => (
```
MichaelRoytman: In order to test this action creator, we need to return a single promise. The [Promise.all](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all) call gathers up each promise per-file and only resolves when all underlying promises resolve.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
       clientApi.postUploadAsset(courseDetails.id, file)
         .then((response) => {
           if (response.ok) {
-            response.json().then((data) => {
-              dispatch(uploadAssetSuccess(data.asset, response));
-              dispatch(getAssets(getState().request, courseDetails));
+            return response.json().then((json) => {
```
MichaelRoytman: In order to correctly chain the promises, we need to return the promise up. See the section on chaining [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Using_promises). 

> Important: Always return promises up, otherwise callbacks won't chain, and errors won't be caught (arrow functions return implicitly when {} are left out).

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
+              dispatch(uploadAssetSuccess(json));
+              // dispatch(getAssets(..)) returns a promise, so we return it
+              // up to chain it
+              return dispatch(getAssets(getState().request, courseDetails));
```
MichaelRoytman: Furthermore, getAssets(...) returns a function, which, when dispatched, returns another promise. Therefore, we also have to return up this promise.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
             });
-          } else {
-            dispatch(uploadAssetFailure(file, response));
           }
-        });
-    });
+          dispatch(uploadAssetFailure(file, response.status));
+          return undefined;
```
MichaelRoytman: I'm a little iffy on this. I was forced to add it because of [this](https://eslint.org/docs/rules/consistent-return) eslint rule: 
> Expected to return a value at the end of arrow function. (consistent return)

Does this seem okay to everyone?

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
+              // dispatch(getAssets(..)) returns a promise, so we return it
+              // up to chain it
+              return dispatch(getAssets(getState().request, courseDetails));
             });
-          } else {
-            dispatch(uploadAssetFailure(file, response));
           }
-        });
-    });
+          dispatch(uploadAssetFailure(file, response.status));
```
MichaelRoytman: This has been moved out of the else case, since the else case was [unnecessary](https://eslint.org/docs/rules/no-else-return):

> Unnecessary 'else' after 'return'. (no-else-return)

src/data/actions/assets.test.jsx
```diff
@@ -138,6 +149,42 @@ describe('Assets Action Creators', () => {
       expect(store.getActions()).toEqual(expectedActions);
     });
   });
+  it('returns expected state from toggleLockAsset success', () => {
```
MichaelRoytman: Farhanah originally reviewed the two toggleLockAsset tests below, but I accidentally removed them when rebasing my previous test branch 😞 . I've tried to reproduce them as closely to the originals as I could remember. I think that they're probably fine to not be reviewed, since they've been looked at, but feel free to take a look if you want to.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
   count,
 });
 
-export const uploadAssetSuccess = (asset, response) => ({
+export const uploadAssetSuccess = response => ({
```
MichaelRoytman: The uploadAssets action creator has been modified to just fire an action with only the JSONified response as the payload, which includes the value of asset, so I've changed this action creator to just take response as a parameter. The reducer that consumes this action can pick out the asset from the response value.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
   type: assetActions.UPLOAD_ASSET_SUCCESS,
-  asset,
   response,
 });
 
-export const uploadAssetFailure = (file, response) => ({
+export const uploadAssetFailure = (asset, response) => ({
```
MichaelRoytman: I've changed file to asset to be more consistent with the way we've been referring to assets in the codebase. If we prefer the term "file", we should change everything to be file consistently.

src/data/actions/assets.js
```diff
@@ -121,46 +121,44 @@ export const uploadingAssets = count => ({
...
+    // gather all the promises into a single promise that can be returned
+    return Promise.all(files.map(file => (
       clientApi.postUploadAsset(courseDetails.id, file)
         .then((response) => {
           if (response.ok) {
-            response.json().then((data) => {
-              dispatch(uploadAssetSuccess(data.asset, response));
-              dispatch(getAssets(getState().request, courseDetails));
+            return response.json().then((json) => {
+              dispatch(uploadAssetSuccess(json));
```
MichaelRoytman: I've changed the uploadAssetSuccess action creator to just take the JSONified response, since the raw Response isn't super useful, and asset happens to be a part of the response JSON anyway.

src/data/actions/assets.js
MichaelRoytman: I've changed `response` to `response.status` just because I don't want to send over the raw Response object. Is there a better alternative? Should I model this after the getAssets action creator, which creates an Error object using the Response and uses that as the action payload? See [here](https://github.com/edx/studio-frontend/blob/master/src/data/actions/assets.js#L35-L45).

src/data/actions/assets.test.jsx
```diff
@@ -159,15 +206,86 @@ describe('Assets Action Creators', () => {
     expect(store.dispatch(actionCreators.uploadExceedMaxCount(1))).toEqual(expectedAction);
   });
   it('returns expected state from uploadAssetFailure', () => {
-    const expectedAction = { file: 'file', response: 'response', type: assetActions.UPLOAD_ASSET_FAILURE };
+    const expectedAction = { asset: 'file', response: 'response', type: assetActions.UPLOAD_ASSET_FAILURE };
```
MichaelRoytman: Just changing the test to match the changes I made to the uploadAssetFailure action creator.

src/data/actions/assets.test.jsx
```diff
@@ -159,15 +206,86 @@ describe('Assets Action Creators', () => {
...
     expect(store.dispatch(actionCreators.uploadAssetFailure('file', 'response'))).toEqual(expectedAction);
   });
   it('returns expected state from uploadAssetSuccess', () => {
-    const expectedAction = { asset: 'asset', response: 'response', type: assetActions.UPLOAD_ASSET_SUCCESS };
-    expect(store.dispatch(actionCreators.uploadAssetSuccess('asset', 'response'))).toEqual(expectedAction);
+    const expectedAction = { response: 'response', type: assetActions.UPLOAD_ASSET_SUCCESS };
+    expect(store.dispatch(actionCreators.uploadAssetSuccess('response'))).toEqual(expectedAction);
```
MichaelRoytman: Just changing the test to match the changes I made to the uploadAssetSuccess action creator.

src/data/actions/assets.test.jsx
```diff
@@ -159,15 +206,72 @@ describe('Assets Action Creators', () => {
...
   });
+  it('returns expected state from uploadAssets success', () => {
+    const assetsResponse = {
+      status: 200,
+      body: {
+        assets: ['a.txt'],
+      },
+    };
+
+    fetchMock.mock(`begin:${assetsEndpoint}`, getUploadResponse(true), { method: 'post' });
```
MichaelRoytman: The uploadAssets action creator makes two types of fetch calls: one post to upload the asset and one get to get the new assets from the server. These two requests should have different responses, so I've set up the fetch-mock to discriminate them based on whether or not they're post or get requests.

MichaelRoytman: getUploadResponse is a currying function, which, given a boolean that represents whether the mocked response represents a success or a failure, returns a function that fetch-mock uses to return a mocked response. This allows me to dynamically mock a response that has the correct filename for each file.

src/data/reducers/assets.js
```diff
@@ -125,7 +125,7 @@ export const status = (state = {}, action) => {
       };
     case assetActions.UPLOAD_ASSET_FAILURE:
       return {
-        file: action.file,
+        asset: action.asset,
```
MichaelRoytman: I'm just changing the UPLOAD_ASSET_FAILURE reducer to reflect the changes I made to the uploadAssetFailure action creator, which renames the file key to asset.

src/data/reducers/utils.test.js
```diff
@@ -22,16 +22,28 @@ describe('Reducer Utils', () => {
...
+
+      const expectedAssets = deepCopyAssets();
+      expectedAssets[0].loadingFields = [assetLoading.LOCK, mockLoadingField];
+
+      expect(addLoadingField(newAssets, 'lynx', mockLoadingField)).toEqual(expectedAssets);
+    });
   });
 
-  describe('addLoadingField', () => {
+  describe('removeLoadingField', () => {
```
MichaelRoytman: Just fixing a typo, since the tests in this describe block are for the removeLoadingField utility function.

src/data/reducers/utils.test.js
```diff
@@ -22,16 +22,28 @@ describe('Reducer Utils', () => {
       expectedAssets[1].loadingFields = [assetLoading.LOCK];
       expect(addLoadingField(assets, 'chinchilla', assetLoading.LOCK)).toEqual(expectedAssets);
     });
+    it('adds loading field to existing loading fields', () => {
```
MichaelRoytman: I've added a test case to cover [this](https://github.com/edx/studio-frontend/blob/master/src/data/reducers/utils.js#L18) line. In particular, it tests the case where an asset already has an existing loadingField and the utility function is called to add another loadingField. I created a dummy loadingField called mockLoadingField to be able to add another type of loadingField to the asset, since we currently only have LOCK in constants/loadingTypes.

src/data/actions/assets.js
fysheets: Should this be updated to use `assets` and `asset` instead of `files/file` as you did with `uploadAssetFailure`?

src/data/actions/assets.js
fysheets: This is done the same way in the accessibility actions file: https://github.com/edx/studio-frontend/blob/master/src/data/actions/accessibility.js#L27

While I'm not 100% on it I think it's fine.

src/data/actions/assets.js
fysheets: I think status is fine for now, once we have a valid use case of error responses (non-2XX) we can revaluate.

src/data/actions/assets.js
fysheets: Bumping this one ^

PR #92, opened 2018-01-16
Move headers to SFE-only template

thallada:
Since we decided to use the Studio headers in EDUCATOR-2067, I'm removing them from the AssetsPage and SF app.

Now the "studio-frontend" header will only appear on the SFE page outside Studio.

@edx/educator-dahlia

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.8.1",
+  "version": "0.8.2",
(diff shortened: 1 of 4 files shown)


thallada:
@fysheets Here's what it looks like in SFE-only:

localhost_18011_assets html

And what it looks like in Studio with my other PR that hides the upload button and side bar text:

localhost_18010_assets_course-v1_edx demox demo_course_

PR #93, opened 2018-01-16
StatusAlert Fixes

fysheets:
# [EDUCATOR-2079](https://openedx.atlassian.net/browse/EDUCATOR-2079)

Updates to ensure the StatusAlert is only visible when necessary, shows the correct data (clears/updates appropriately). This brings in upload count & StatusAlert context fields into the component state.

### DIFF
--- a/src/components/AssetsTable/index.jsx
+++ b/src/components/AssetsTable/index.jsx
@@ -19,12 +19,18 @@ export class AssetsTable extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      modalOpen: false,
       assetToDelete: {},
+      copyButtonIsClicked: false,
       deletedAsset: {},
       deletedAssetIndex: null,
       elementToFocusOnModalClose: {},
-      copyButtonIsClicked: false,
+      modalOpen: false,
+      statusAlertFields: {
+        alertDialog: '',
+        alertType: 'info',
+      },
+      statusAlertOpen: false,
+      uploadSuccessCount: 1,
     };
 
     this.columns = {
@@ -68,16 +74,25 @@ export class AssetsTable extends React.Component {
       },
     };
 
+    this.deleteActions = [assetActions.DELETE_ASSET_FAILURE, assetActions.DELETE_ASSET_SUCCESS];
+
     this.trashcanRefs = {};
     this.statusAlertRef = {};
 
-    this.onDeleteClick = this.onDeleteClick.bind(this);
-    this.deleteAsset = this.deleteAsset.bind(this);
     this.addSupplementalTableElements = this.addSupplementalTableElements.bind(this);
     this.closeModal = this.closeModal.bind(this);
     this.closeStatusAlert = this.closeStatusAlert.bind(this);
-    this.renderStatusAlert = this.renderStatusAlert.bind(this);
+    this.deleteAsset = this.deleteAsset.bind(this);
     this.onCopyButtonClick = this.onCopyButtonClick.bind(this);
+    this.onDeleteClick = this.onDeleteClick.bind(this);
+    this.renderStatusAlert = this.renderStatusAlert.bind(this);
+    this.updateUploadSuccessCount = this.updateUploadSuccessCount.bind(this);
+  }
+
+
+  componentWillReceiveProps(nextProps) {
+    const { assetsStatus } = nextProps;
+    this.updateStatusAlertFields(assetsStatus);
   }
 
   onSortClick(columnKey) {
@@ -173,68 +188,6 @@ export class AssetsTable extends React.Component {
     />);
   }
 
-  getLockFailedStatusFields(assetsStatus) {
-    return {
-      alertDialog: `Failed to toggle lock for ${assetsStatus.asset.name}`,
-      alertType: 'danger',
-    };
-  }
-
-  getDeleteStatusFields(assetsStatus) {
-    const assetName = this.state.deletedAsset.display_name;
-    let alertDialog;
-    let alertType;
-
-    switch (assetsStatus.type) {
-      case assetActions.DELETE_ASSET_FAILURE:
-        alertDialog = `Unable to delete ${assetName}.`;
-        alertType = 'danger';
-        break;
-      case assetActions.DELETE_ASSET_SUCCESS:
-        alertDialog = `${assetName} has been deleted.`;
-        alertType = 'success';
-        break;
-      default:
-        break;
-    }
-    return {
-      alertDialog,
-      alertType,
-    };
-  }
-
-  getUploadStatusFields(assetsStatus) {
-    switch (assetsStatus.type) {
-      case assetActions.UPLOAD_ASSET_SUCCESS:
-        return {
-          alertType: 'success',
-          alertDialog: `${assetsStatus.loadedCount} files successfully uploaded.`,
-        };
-      case assetActions.UPLOADING_ASSETS:
-        return {
-          alertType: 'info',
-          alertDialog: `${assetsStatus.count} files uploading.`,
-        };
-      case assetActions.UPLOAD_EXCEED_MAX_COUNT_ERROR:
-        return {
-          alertType: 'danger',
-          alertDialog: `The maximum number of files for an upload is ${assetsStatus.maxFileCount}. No files were not uploaded.`,
-        };
-      case assetActions.UPLOAD_EXCEED_MAX_SIZE_ERROR:
-        return {
-          alertType: 'danger',
-          alertDialog: `The maximum size for an upload is ${assetsStatus.maxFileSizeMB} MB. No files were not uploaded.`,
-        };
-      case assetActions.UPLOAD_ASSET_FAILURE:
-        return {
-          alertType: 'danger',
-          alertDialog: `Error uploading ${assetsStatus.file.name}. Try again.`,
-        };
-      default:
-        return {};
-    }
-  }
-
   getCopyUrlButtons(assetDisplayName, studioUrl, webUrl) {
     return (
       <span>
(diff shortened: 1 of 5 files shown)


src/components/AssetsTable/index.jsx
```diff
@@ -19,12 +19,18 @@ export class AssetsTable extends React.Component {
...
       elementToFocusOnModalClose: {},
-      copyButtonIsClicked: false,
+      modalOpen: false,
+      statusAlertFields: {
+        alertType: 'info',
+        alertDialog: '',
+      },
+      statusAlertOpen: false,
+      uploadSuccessCount: 1,
     };
```
fysheets: Alphabetizing & adding statusAlert fields + uploadSuccessCount

src/components/AssetsTable/index.jsx
```diff
@@ -68,16 +74,25 @@ export class AssetsTable extends React.Component {
...
     this.closeModal = this.closeModal.bind(this);
     this.closeStatusAlert = this.closeStatusAlert.bind(this);
-    this.renderStatusAlert = this.renderStatusAlert.bind(this);
+    this.deleteAsset = this.deleteAsset.bind(this);
     this.onCopyButtonClick = this.onCopyButtonClick.bind(this);
+    this.onDeleteClick = this.onDeleteClick.bind(this);
+    this.renderStatusAlert = this.renderStatusAlert.bind(this);
+    this.updateUploadSuccessCount = this.updateUploadSuccessCount.bind(this);
+  }
+
```
fysheets: Alphabetizing + updateUploadSuccesCount

src/components/AssetsTable/index.jsx
```diff
@@ -68,16 +74,25 @@ export class AssetsTable extends React.Component {
...
+
+  componentWillReceiveProps(nextProps) {
+    const { assetsStatus } = nextProps;
+    this.getStatusAlertFields(assetsStatus);
   }
```
fysheets: Will update StatusAlertFields if the incoming props denote an update is needed for rendering the StatusAlert

src/components/AssetsTable/index.jsx
```diff
@@ -68,16 +74,25 @@ export class AssetsTable extends React.Component {
       },
     };
 
+    this.deleteActions = [assetActions.DELETE_ASSET_FAILURE, assetActions.DELETE_ASSET_SUCCESS];
+
```
fysheets: Would it make sense to format these action types into delete/upload/lock/etc within the constants file [here](https://github.com/edx/studio-frontend/blob/master/src/data/constants/actionTypes.js#L3) instead? Something like:

`export const assetActions = {
  requestAssets: {
    REQUEST_ASSETS_SUCCESS: 'REQUEST_ASSETS_SUCCESS',
    REQUEST_ASSETS_FAILURE: 'REQUEST_ASSETS_FAILURE',
  },
  deleteAssets: {
    DELETE_ASSET_SUCCESS: 'DELETE_ASSET_SUCCESS',
    DELETE_ASSET_FAILURE: 'DELETE_ASSET_FAILURE',
  },
}`

And if we need to check if the assetType is a part of a subset (like we do [here](https://github.com/edx/studio-frontend/pull/93/files#diff-35e343ee67d70894218ba932e51bd1cdR412) ) then we won't have to replicate that data/update it if we add more types?

src/components/AssetsTable/index.jsx
```diff
@@ -19,12 +19,18 @@ export class AssetsTable extends React.Component {
...
       assetToDelete: {},
+      copyButtonIsClicked: false,
       deletedAsset: {},
       deletedAssetIndex: null,
       elementToFocusOnModalClose: {},
-      copyButtonIsClicked: false,
+      modalOpen: false,
+      statusAlertFields: {
+        alertType: 'info',
+        alertDialog: '',
```
MichaelRoytman: Nit: Should we alphabetize the fields of this object too?

src/components/AssetsTable/index.jsx
MichaelRoytman: I think that's a good idea. Would there be any reason to keep the constants file as-is (i.e. would categorizing the action types this way make it more difficult to access them for other components/files?).

efischer19: I'm on board with this

src/components/AssetsTable/index.jsx
```diff
@@ -270,6 +223,67 @@ export class AssetsTable extends React.Component {
...
+        alertDialog = `${this.state.uploadSuccessCount} files successfully uploaded.`;
+        break;
+      case assetActions.UPLOADING_ASSETS:
+        this.closeStatusAlert();
+        alertType = 'info';
+        alertDialog = `${assetsStatus.count} files uploading.`;
+        break;
+      case assetActions.UPLOAD_EXCEED_MAX_COUNT_ERROR:
+        alertType = 'danger';
+        alertDialog = `The maximum number of files for an upload is ${assetsStatus.maxFileCount}. No files were not uploaded.`;
```
MichaelRoytman: 'No files were not uploaded' sounds a little funny. I think you don't need the extra 'not'. What do you think?

src/components/AssetsTable/index.jsx
```diff
@@ -270,6 +223,67 @@ export class AssetsTable extends React.Component {
...
+        break;
+      case assetActions.UPLOAD_EXCEED_MAX_SIZE_ERROR:
+        alertType = 'danger';
+        alertDialog = `The maximum size for an upload is ${assetsStatus.maxFileSizeMB} MB. No files were not uploaded.`;
```
MichaelRoytman: There's the extra 'not' here as well.

src/components/AssetsTable/index.jsx
```diff
@@ -270,6 +223,67 @@ export class AssetsTable extends React.Component {
...
+        break;
+      case assetActions.UPLOAD_ASSET_FAILURE:
+        alertType = 'danger';
+        alertDialog = `Error uploading ${assetsStatus.file.name}. Try again.`;
+        break;
+      case assetActions.TOGGLING_LOCK_ASSET_FAILURE:
+        alertType = 'danger';
+        alertDialog = `Failed to toggle lock for ${assetsStatus.asset.name}`;
```
MichaelRoytman: Nit: Could we add a period to the end of this sentence to be consistent with the other alert messages?

src/components/AssetsTable/index.jsx
```diff
@@ -332,6 +346,12 @@ export class AssetsTable extends React.Component {
     this.setState({
       deletedAsset: {},
       deletedAssetIndex: null,
+      statusAlertOpen: false,
```
MichaelRoytman: I'm not sure that this suggestion is better - just something I thought of. Would it be worth it to have some kind of variable 'defaultStatusAlertState' that we could refer to here? That way, we could just replace these lines with { ...defaultStatusAlertState}, and use that in the constructor as well. I'm just a little worried it won't be super clear that we're just resetting back to the defaults in this setState call. The use of the comment just makes me wonder if we could try to be a little bit more clear here. Maybe even pulling the setState out into a function 'resetStatusState'? I don't feel super strongly about it - just throwing it out there! :)

src/components/AssetsTable/index.jsx
```diff
@@ -386,44 +405,19 @@ export class AssetsTable extends React.Component {
...
-      assetActions.UPLOAD_EXCEED_MAX_COUNT_ERROR, assetActions.UPLOAD_EXCEED_MAX_SIZE_ERROR,
-      assetActions.UPLOAD_ASSET_FAILURE, assetActions.UPLOADING_ASSETS];
-    let status = {};
+
     let onClose = this.closeStatusAlert;
-    if (assetsStatus.type === assetActions.TOGGLING_LOCK_ASSET_FAILURE) {
-      status = this.getLockFailedStatusFields(assetsStatus);
-    } else if (deleteActions.includes(assetsStatus.type)) {
-      status = this.getDeleteStatusFields(assetsStatus);
+    if (this.deleteActions.includes(assetsStatus.type)) {
```
MichaelRoytman: I know that we always make a getAssets call on every page load, so there is always some kind of assetsStatus state. But the default assetsStatus state is {}. Is there any reason we'd get into a position where assetsStatus is {}? If our first getAssets times out, would it remain {}? I only ask because then assetsStats.type would be undefined, and I'm not sure what would happen in that case. Should we add something to be a little more robust to that? Or should we just let it go, since we haven't run into that yet?

MichaelRoytman: I was just looking at how Array.includes works, and I saw that it's [not supported](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) by Internet Explorer :(. It seems to me that babel has a [polyfill](https://github.com/zloirock/core-js#stage-4-proposals) in the babel-polyfill pacakage, so I think if you do `import "babel-polyfill";` at the top of the file, that should take care of it. I found that [here](https://babeljs.io/docs/usage/polyfill/). I think there might be some other places in the codebase where we use Array.includes, so we should polyfill it there. I'm not sure whether that's critical to do now, or whether we can file a ticket to do later. I know Internet Explorer isn't used super often, but I'd hate to release something and then have someone use Internet Explorer and see it break :/. I'd be interested in other people's opinions on that, though.

src/components/AssetsTable/index.jsx
fysheets: Yes :)

src/components/AssetsTable/index.jsx
fysheets: I don't think it would be more difficult, but it may require additional updates in other files in this PR to handle the refactor correctly. The change would be - `assetActions.DELETE_ASSET_SUCCESS` to `assetActions.deleteAssets.DELETE_ASSET_SUCCESS` - so that would be more verbose / more to remember when referencing things / slightly duplicative.  Maybe breaking it apart by API type instead of by page? `deleteActions.DELETE_ASSET_SUCCESS`? We can hash this out tomorrow in person too.

src/components/AssetsTable/index.jsx
fysheets: Ah that must have been carried over, I'll remove the 'not'.

src/components/AssetsTable/index.jsx
fysheets: If we go the route of separating out the `actionTypes` above to more distinct types we should be able to leverage `if ('key' in myObj)` for logical checks like this as we're be checking if it's in an `Object` instead of an `Array`. I think this kind of makes me want to push to organizing those action types more distinctly so that we don't need to use `includes`.

fysheets: See comment below -- I would want to sanity check the `undefined`/`null` cases there.

src/components/AssetsTable/index.jsx
fysheets: Would it be enough to rename the existing function to `clearStatusAlert`?

src/components/AssetsTable/index.jsx
fysheets: Also - We can merge this in as is with the Array (noting the Array.includes issue from below) and I'll create an immediate follow up branch to do the reactor of the action types so that this PR doesn't balloon.

src/components/AssetsTable/index.jsx
fysheets: Update: calling `hasOwnProperty()` with `null` or `undefined` simply returns `false`.  I tested with the built in area here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty. Here's the link for the `in` operator as well for reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/in - it looks like we can leverage `in` with Arrays as well, but I think in this case, updating to use the constants objects if we can as well as `hasOwnProperty` to validate on the top level is the best route if we make a change here.

src/components/AssetsTable/index.jsx
MichaelRoytman: Looking at this again, I think the function clears and closes, so I think it's probably fine as is, in my opinion. Thanks for entertaining the idea, though.

fysheets:
@MichaelRoytman if this looks good to you now, I'll do a fix version bump on top of it once your review has updated :)

PR #96, opened 2018-01-18
fix sr-only style so Copied does not appear in browser

MichaelRoytman:


MichaelRoytman:
@fysheets Thanks for the review. I tried rebasing on top of master to see if being behind was affecting the coverage. I'll see if it moves the coverage back up or not. If not, I'll just merge it. Codecov is kind of difficult to follow sometimes.

PR #97, opened 2018-01-19
Implement eslint browser compatibility

jaebradley:
The [`eslint-plugin-compat`](https://github.com/amilajack/eslint-plugin-compat) dependency enables linting for various browser versions to avoid accidentally using features or syntax that is not consistent with the set of browsers and versions you want to support.

The current browser / version combination is the past two versions of all browsers except for IE which is required to be 11 or higher. Feel free to adjust this configuration.

I've also `eslint-ignore`d existing compatibility errors.

@edx/educator-dahlia

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -14,9 +14,12 @@
     ],
     "class-methods-use-this": [
       "off"
-    ]
+    ],
+    "compat/compat": "error"
   },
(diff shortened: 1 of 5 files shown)


fysheets:
Closing this out for now as this repo is no longer in active development.

PR #99, opened 2018-01-23
Update page value when filtering assets

fysheets:
# [EDUCATOR-2182](https://openedx.atlassian.net/browse/EDUCATOR-2182)

Update to ensure on every filter we reset to the first page in the list of assets being fetched prior to fetching them. This is also related to [EDUCATOR-2182](https://openedx.atlassian.net/browse/EDUCATOR-2182) but should be validated to resolve both issues during review.

FYI @edx/educator-dahlia

### DIFF
--- a/src/components/AssetsFilters/AssetsFilters.test.jsx
+++ b/src/components/AssetsFilters/AssetsFilters.test.jsx
@@ -1,12 +1,14 @@
 import React from 'react';
 import Enzyme from 'enzyme';
-import { AssetsFilters } from './index';
+import { AssetsFilters, mapDispatchToProps } from './index';
+import { assetActions } from '../../data/constants/actionTypes';
 
 const mount = Enzyme.mount;
 
 const defaultProps = {
   assetsFilters: {},
   updateFilter: () => {},
+  updatePage: () => {},
 };
 
 let wrapper;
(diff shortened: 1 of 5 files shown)


fysheets:
Thanks @thallada will do!

src/components/AssetsFilters/AssetsFilters.test.jsx
```diff
@@ -39,5 +41,56 @@ describe('<AssetsFilters />', () => {
...
+        updateFilter: filterSpy,
+        updatePage: pageUpdateSpy,
+      });
+
+      const checkBoxGroup = wrapper.find('CheckBoxGroup');
+      const checkBoxes = checkBoxGroup.find('[type="checkbox"]');
+      const checkBox = checkBoxes.first();
+      checkBox.simulate('change', { target: { checked: true, type: 'checkbox' } });
+
+      expect(pageUpdateSpy).toHaveBeenCalledTimes(1);
```
MichaelRoytman: The two tests look good. The only thing I can think of is also adding something like:

`expect(pageUpdateSpy).toHaveBeenCalledWith(0)` to make sure that we're actually resetting the page back to 0 when a filter is clicked. Does that seem reasonable?

I'd suggest the same thing for filterSpy, but I'm not sure what the filter is off the top of my head. I'd imagine something like: 

*click filter for Images*
*expect filterSpy to be called with { Images: true }, or whatever it is that it's called with*

fysheets: That does make sense and sounds reasonable -- thanks!

PR #101, opened 2018-01-25
Wrap text in assets table.

dsjen:
**Before**
before
* long file names didn't wrap -- see the cut off column to the right of the screenshot
---

**After**
after
* long file names will wrap
* wrapped words in this case are arbitrarily wrapped--see file name and file type columns

---
Another option entirely would be to use an ellipsis, but it might be difficult to distinguish between file names that start off the same.

@edx/educator-dahlia -- I ended up needing to go with `break-all` as FF wasn't breaking with `break-word`.

@sstack22 -- please confirm this is ok from a product perspective in addressing table sizing issues.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.8.6",
+  "version": "0.8.7",
   "description": "The frontend for the Open edX platform",
   "repository": "edx/studio-frontend",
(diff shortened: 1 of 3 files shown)


sstack22:
👍 , looks good!

dsjen:
Great, thanks! I'm going to bump the patch version in this PR and will merge once that's set.

PR #103, opened 2018-01-25
mroytman/add no results view

MichaelRoytman:


src/components/AssetsFilters/index.jsx
```diff
@@ -5,31 +5,9 @@ import CheckBox from '@edx/paragon/src/CheckBox';
 import { connect } from 'react-redux';
 
 import { filterUpdate, pageUpdate } from '../../data/actions/assets';
+import ASSET_TYPES from '../../data/constants/assetTypeFilters';
 import styles from './AssetsFilters.scss';
 
-const ASSET_TYPES = [
```
MichaelRoytman: Since these are constants, I have moved them into /data/constants/assetTypeFilters.

src/components/AssetsPage/pageTypes.jsx
```diff
@@ -0,0 +1,79 @@
...
+  NO_RESULTS: 'noResults',
+};
+
+const renderAssetsDropZone = maxFileSizeMB => (
+  <WrappedAssetsDropZone
+    maxFileSizeMB={maxFileSizeMB}
+  />
+);
+
+export const renderAssetsPage = maxFileSizeMB => (
```
MichaelRoytman: This is the normal assets page.

src/components/AssetsPage/pageTypes.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+const renderNoAssetsBody = () => (
+  <React.Fragment>
+    <h3>0 files in your course</h3>
+    <h4>Enhance your course content by uploading files such as images and documents.</h4>
+  </React.Fragment>
+
+);
+
+export const renderNoAssetsPage = maxFileSizeMB => (
```
MichaelRoytman: This is the assets page when there are no assets in the course.

src/components/AssetsPage/pageTypes.jsx
```diff
@@ -0,0 +1,79 @@
...
+    <h4>{ numberOfFilters > 1 ? 'No files were found for these filters.' : 'No files were found for this filter.'}</h4>
+    <Button
+      buttonType="link"
+      onClick={clearFilters}
+      label={numberOfFilters > 1 ? 'Clear all filters' : 'Clear filter'}
+    />
+  </React.Fragment>);
+};
+
+export const renderNoResultsPage = (maxFileSizeMB, request, clearFilters) => (
```
MichaelRoytman: This is the assets page when the filter results are empty.

src/components/AssetsPage/index.jsx
```diff
@@ -44,63 +37,41 @@ export class AssetsPage extends React.Component {
...
+    return pageTypes.types.NO_RESULTS;
+  }
 
-  hasAssets = assetsList => (
-    assetsList.length > 0
-  )
+  getPage = (type) => {
+    switch (type) {
+      case pageTypes.types.NORMAL:
+        return pageTypes.renderAssetsPage(this.props.uploadSettings.max_file_size_in_mbs);
```
MichaelRoytman: I'm a little iffy on this part. Originally, I just passed all of props, but then Farhanah and I settled on just passing in the necessary parameters. Just flagging this as a potential "code smell". What happens if these pages get more complex and require more data from the AssetsPage?

src/components/AssetsPage/index.jsx
```diff
@@ -44,63 +37,41 @@ export class AssetsPage extends React.Component {
...
+    }
+  }
 
   handleCheckBoxChange = (checked) => {
     this.setState({
       checked,
     });
   }
 
-  renderAssetsDropZone = () => (
```
MichaelRoytman: These functions have been moved to AssetsPage/pageTypes.jsx.

src/components/AssetsPage/index.jsx
```diff
@@ -133,13 +106,14 @@ AssetsPage.propTypes = {
 };
 
 const mapStateToProps = state => ({
+  assetsList: state.assets,
```
MichaelRoytman: Just alphabetizing!

src/data/reducers/assets.test.jsx
```diff
@@ -313,7 +318,7 @@ describe('Assets Reducers', () => {
         ...action.data,
       });
     });
-    it('returns correct sort and direction state on PAGE_UPDATE action', () => {
+    it('returns correct page state on PAGE_UPDATE action', () => {
```
MichaelRoytman: Fixing old copy-paste mistake.

src/data/reducers/assets.test.jsx
```diff
@@ -328,7 +333,7 @@ describe('Assets Reducers', () => {
         ...action.data,
       });
     });
-    it('returns correct sort and direction state on FILTER_UPDATE action', () => {
+    it('returns correct assetTypes state on FILTER_UPDATE action', () => {
```
MichaelRoytman: Fixing old copy-paste mistake.

src/data/reducers/assets.test.jsx
```diff
@@ -345,5 +350,23 @@ describe('Assets Reducers', () => {
         assetTypes: { ...defaultState.assetTypes, ...action.data },
       });
     });
+    it('returns correct assetTypes state on CLEAR_FILTERS action', () => {
```
MichaelRoytman: Testing new CLEAR_FILTERS action.

package.json
```diff
@@ -13,7 +13,7 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^0.4.3",
-    "@edx/paragon": "^1.7.1",
+    "@edx/paragon": "^2.0.0",
```
MichaelRoytman: CheckBox clearing.

MichaelRoytman:
@edx/educator-dahlia Please review!

src/components/AssetsFilters/AssetsFilters.test.jsx
```diff
@@ -2,13 +2,15 @@ import React from 'react';
 import Enzyme from 'enzyme';
 import { AssetsFilters, mapDispatchToProps } from './index';
 import { assetActions } from '../../data/constants/actionTypes';
+import { filtersInitial } from './../../data/reducers/assets';
 
 const mount = Enzyme.mount;
 
 const defaultProps = {
-  assetsFilters: {},
+  assetsFilters: { ...filtersInitial },
```
MichaelRoytman: I have changed the initial defaultProps for the AssetsFilters component to be filtersInitial, which is

{
   Audio: false,
   Code: false,
   Documents: false,
   Images: false,
   Other: false,
}

src/components/AssetsFilters/index.jsx
```diff
@@ -41,7 +19,7 @@ export const AssetsFilters = ({ assetsFilters, updateFilter, updatePage }) => (
             id={type.key}
             name={type.key}
             label={type.displayName}
-            checked={assetsFilters[type.key]}
+            checked={assetsFilters.assetTypes[type.key]}
```
MichaelRoytman: assetsFilters is an object with a key assetTypes, which is a dictionary of key-value pairs "filter": "is-filter-enabled". This is so that we can leave room for other kinds of filters in the future.

src/components/AssetsPage/index.jsx
```diff
@@ -44,63 +37,41 @@ export class AssetsPage extends React.Component {
     }
   }
 
-  hasNoAssets = assetsList => (
-    assetsList.length === 0
-  )
+  getPageType = (props) => {
+    if (props.assetsList.length > 0) {
+      return pageTypes.types.NORMAL;
+    } else if (Object.keys(props.request.assetTypes).length === 0) {
```
MichaelRoytman: This condition assumes props.request.assetTypes exists. Should I make this more robust by checking `'assetTypes' in props.request` first? The default request state in the reducer is the assetTypes object, so that key always exists, but only when the React component is tied to our particular Redux state. Just wanted to check!

src/components/AssetsPage/index.jsx
```diff
@@ -109,7 +80,9 @@ export class AssetsPage extends React.Component {
 }
 
 AssetsPage.propTypes = {
+  // eslint-disable-next-line react/no-unused-prop-types
```
MichaelRoytman: I don't think eslint is smart enough to realize that I actually am using the assetsList prop at line 41 of this file, so I have disabled the rule for that reason.

src/components/AssetsPage/pageTypes.jsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import Button from '@edx/paragon/src/Button';
+
+import WrappedAssetsDropZone from '../AssetsDropZone';
+import WrappedAssetsTable from '../AssetsTable';
+import WrappedAssetsFilters from '../AssetsFilters';
+import WrappedPagination from '../Pagination';
+import edxBootstrap from '../../SFE.scss';
+
+export const types = {
```
MichaelRoytman: I'm not married to these enums. I'm happy to change them to something else if there are other opinions for a better naming scheme.

src/components/AssetsPage/pageTypes.jsx
```diff
@@ -0,0 +1,79 @@
...
+  <WrappedAssetsDropZone
+    maxFileSizeMB={maxFileSizeMB}
+  />
+);
+
+export const renderAssetsPage = maxFileSizeMB => (
+  <React.Fragment>
+    <div className={edxBootstrap.col}>
+      { renderAssetsDropZone(maxFileSizeMB) }
+      <WrappedAssetsFilters />
```
MichaelRoytman: TODO: I should probably pull this out into a renderAssetsFilters function like the renderAssetsDropZone function above it.

src/components/AssetsPage/pageTypes.jsx
```diff
@@ -0,0 +1,79 @@
...
+      { renderAssetsDropZone(maxFileSizeMB) }
+    </div>
+    <div className={edxBootstrap['col-10']}>
+      { renderNoAssetsBody() }
+    </div>
+  </React.Fragment>
+);
+
+const renderNoResultsBody = (request, clearFilters) => {
+  const numberOfFilters = Object.keys(request.assetTypes)
```
MichaelRoytman: This returns an array containing the filter types that are enabled.

src/components/AssetsPage/index.jsx
```diff
@@ -1,41 +1,34 @@
...
 export class AssetsPage extends React.Component {
   constructor(props) {
     super(props);
 
-    const hasAssets = this.hasAssets(props.assetsList);
-
     this.state = {
-      checked: false,
-      hasAssets,
+      pageType: pageTypes.types.SKELETON,
```
MichaelRoytman: The initial page type is the SKELETON page. The page type changes as soon as the component receives new props (e.g. a new assets list). This is why, in the tests, I have to always call setProps to trigger a page change. If a page is initialized with data, it will still be the SKELETON page. The reason for this is what I can't call the getPageType function from the constructor, since it relies on reading the state. Is this okay with you guys?

src/components/AssetsPage/index.jsx
```diff
@@ -44,63 +37,41 @@ export class AssetsPage extends React.Component {
...
 
-  hasAssets = assetsList => (
-    assetsList.length > 0
-  )
-
-  handleCheckBoxChange = (checked) => {
-    this.setState({
-      checked,
-    });
+    if ('type' in props.status && props.status.type === assetActions.request.REQUESTING_ASSETS) {
```
MichaelRoytman: I have added a REQUESTING_ASSETS action. This represents the time between making an API call and getting the response. This was added to prevent the page type from changing in between those two periods. The logic here is that, if we've made a request and are waiting on the response, do not change the page view until the response is received, so that user's experience is not interrupted.

src/components/AssetsPage/index.jsx
```diff
@@ -44,63 +37,41 @@ export class AssetsPage extends React.Component {
...
+        return pageTypes.renderAssetsPage(this.props.uploadSettings.max_file_size_in_mbs);
+      case pageTypes.types.NO_ASSETS:
+        return pageTypes.renderNoAssetsPage(this.props.uploadSettings.max_file_size_in_mbs);
+      case pageTypes.types.NO_RESULTS:
+        return pageTypes.renderNoResultsPage(this.props.uploadSettings.max_file_size_in_mbs,
+          this.props.filtersMetaData, this.clearFilters);
+      case pageTypes.types.SKELETON:
+        return pageTypes.renderSkeletonPage(this.props.uploadSettings.max_file_size_in_mbs);
+      default:
+        return pageTypes.renderSkeletonPage(this.props.uploadSettings.max_file_size_in_mbs);
```
MichaelRoytman: The default is the SKELETON page. Does that seem reasonable? Or should it be the previous page type (à la line 43 of this file)?

src/utils/getAssetsFilters.jsx
```diff
@@ -0,0 +1,23 @@
+import ASSET_TYPES from '../data/constants/assetTypeFilters';
```
MichaelRoytman: This file just has some utility functions for dealing with the filter state.

src/utils/getAssetsFilters.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+export const getSelectedFilters = (filters) => {
```
MichaelRoytman: This returns an array of the filters keys that are selected (i.e. whose values in the filters object are true).

src/utils/getAssetsFilters.jsx
```diff
@@ -0,0 +1,23 @@
...
+  }
+  return Object.keys(filters)
+    .filter(type => filters[type] === true);
+};
+
+export const hasSelectedFilters = filters => (
+  getSelectedFilters(filters).length > 0
+);
+
+export const getDefaultFilterState = () => (
```
MichaelRoytman: This returns the default initial filter state for the filter reducer. It is an object, where the keys are all the asset filters and the values are all false.

src/components/AssetsPage/index.jsx
thallada: If we ever added, for example, server-side rendering, we might have to change this because then the page would actually be rendered once with all of the data and no ajax calls. But, since that's not planned I think this is okay.

src/components/AssetsPage/index.jsx
thallada: AssetsPage isn't much of a component without the pageTypes. I wonder if it would make things simpler to just have those render*Page functions defined on the AssetsPage class so they have access to whatever props that they need via `this.props`.

src/components/AssetsPage/index.jsx
thallada: It doesn't seem like this component would know about the previous page type at this point. It only knows about the current `this.state.pageType` which is passed as `type` to this function. Since the constructor sets that to SKELETON, I don't think this default case should ever be hit.

src/data/actions/assets.js
```diff
@@ -20,9 +20,14 @@ export const requestAssetsFailure = response => ({
...
 
+export const requestingAssets = () => ({
+  type: assetActions.request.REQUESTING_ASSETS,
+});
+
 export const getAssets = (request, courseDetails) =>
-  (dispatch, getState) =>
-    clientApi.requestAssets(courseDetails.id, {
+  (dispatch, getState) => {
+    dispatch(requestingAssets());
```
thallada: I think this makes sense :+1: we can later check for this state and display some sort of loading spinner too.

src/components/AssetsPage/index.jsx
MichaelRoytman: @thallada That's actually what I had originally, but there was some concern about the number of functions in AssetsPage (all of pageTypes.jsx was in index.jsx). I'm happy to move it back, as it would live closer to the data it needs, but I'm not sure how to prevent the file from being enormous. Do you have a preference? I'd actually prefer to move it back, but if there are good arguments for separating it out, I'm happy to keep it.

src/components/AssetsPage/index.jsx
MichaelRoytman: Okay, sounds good!

src/components/AssetsPage/index.jsx
MichaelRoytman: All right. Should the default case just return an empty Object? I like having the case for the SKELETON page be explicit, but I realize it could just be made the default case instead.

thallada: Hmm an error would make the most sense here since it's a case we don't expect.

```
throw new Error(`Unknown pageType {type}`);
```

src/components/AssetsPage/index.jsx
thallada: I somewhat prefer to have the functions in this class, but I can see the argument for separating them too :man_shrugging: . I think it doesn't matter too much either way.

dsjen:
Hooray! It might be on your radar already, but if not, please release this
for the platform. 😃

PR #104, opened 2018-01-25
Remove fsevents and enforce package-lock updates

efischer19:
testing travis stuff for funsies


.travis.yml
```diff
@@ -8,6 +8,7 @@ services:
  - docker
 
 before_script:
+ - git diff --name-only --exit-code
```
efischer19: This runs after the initial `npm install`, and will fail out if there are any changes to checked-in files (read: `package-lock.json`)

PR #106, opened 2018-01-26
EDUCATOR-2207 Scoped CSS

thallada:
This is the first part to fixing our CSS issues in studio-frontend.

1. Scope all CSS (including CSS we import like Bootstrap) to the `<div class="SFE">`
  * This means I will need to make a second PR in platform to add the SFE class to the root div, and to also update the webpack config there since we currently build studio-frontend in the platform.
2. Use the Paragon themeable build so that we don't have to build it ourselves.
3. Don't import all of Bootstrap in each component scss, only the variables and mixins.
  * Now the weight of our CSS is a lot lower because we aren't including the Bootstrap CSS repeatedly for each component.
  * Because of this, some of the `@exports` we used don't work anymore (like for buttons), so instead I reference the Bootstrap classname directly in the JSX. This is safe to do now because we aren't using CSS modules that add a bunch of names and hashes to the classnames.

@edx/educator-dahlia

### DIFF
--- a/config/webpack.dev.config.js
+++ b/config/webpack.dev.config.js
@@ -2,7 +2,6 @@
 
 const Merge = require('webpack-merge');
 const path = require('path');
-const webpack = require('webpack');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
 
 const apiEndpoints = require('../src/data/api/endpoints.js');
@@ -36,7 +35,6 @@ module.exports = Merge.smart(commonConfig, {
         test: /\.(js|jsx)$/,
         include: [
           path.resolve(__dirname, '../src'),
-          path.resolve(__dirname, '../node_modules/@edx/paragon'),
         ],
         loader: 'babel-loader',
         options: {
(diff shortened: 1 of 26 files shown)


config/webpack.dev.config.js
```diff
@@ -87,7 +98,6 @@ module.exports = Merge.smart(commonConfig, {
       filename: 'accessibilityPolicy.html',
       template: path.resolve(__dirname, '../public/index.html'),
     }),
-    new webpack.HotModuleReplacementPlugin(),
```
efischer19: why is this being taken out?

thallada: I think I remember webpack complaining about this. Because [our npm script "start"](https://github.com/edx/studio-frontend/blob/master/package.json#L10) has `--hot` it already automatically adds this plugin. So, this line is redundant.

thallada:
Since this is a backwards incompatible change (platform now has to add `class="SFE"` to the root div) I'm going to make this a major version bump. 1.0 here we go :confetti_ball:

PR #107, opened 2018-01-29
Add make restart command and make restart-detached command.

MichaelRoytman:
…ersonal make commands

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,6 +39,14 @@ from-scratch: ## start development environment from scratch
 	docker build -t edxops/studio-frontend:latest --no-cache .
 	make up
 
+restart:
+	make down
+	make up
+
+restart-detached:
+	make down
+	make up-detached
+
 devstack-install: validate-devstack-folders ## install local version of package into docker devstack for development
 	# note that this assumes a structure /src/sfe_parent/studio-frontend
 	# step 0: remove studio-frontend from edx-platform/node_modules


.gitignore
```diff
@@ -5,3 +5,7 @@ npm-debug.log
 coverage
 
 dist/
+
+# Personal make commands
+local.mk
```
fysheets: Would any of the personal commands that you're adding make sense to be in the overall Makefile? That way we don't have to worry about ignoring them and others could leverage them as well? Or are they very specific to your working environment?

Makefile
```diff
@@ -39,6 +41,10 @@ from-scratch: ## start development environment from scratch
 	docker build -t edxops/studio-frontend:latest --no-cache .
 	make up
 
+restart:
```
fysheets: 👍

.gitignore
MichaelRoytman: I added a `make restart-detached` command for myself, which basically just does `up-detached` after the down. I didn't get the sense that anyone really used the up-detached command, so I didn't want to clutter the Makefile with my commands. If people would use it, I'm happy to add it. What do you think?

fysheets: I think it's a pretty useful command that would be good to have available for the group in general :) One more command won't clutter it up.

MichaelRoytman:
@fysheets I've added the `make restart-detached` make command and removed the .gitignore changes. Does this look okay?

PR #108, opened 2018-01-30
bump version of studio-frontend

MichaelRoytman:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "0.8.7",
+  "version": "1.0.0",
```
efischer19: Why the major bump, as opposed to a minor/patch release?

fysheets: +1 here -- I guess I'm forgetting what the PR was that is related to this bump

PR #109, opened 2018-01-31
EDUCATOR-2225: Deleting File on Assets Page Yields Unexpected Behavior

MichaelRoytman:


package.json
```diff
@@ -77,7 +77,8 @@
       "src/**/*.{js,jsx}"
     ],
     "coveragePathIgnorePatterns": [
-      "/node_modules/"
+      "/node_modules/",
+      "<rootDir>/src/.*/container.jsx"
```
MichaelRoytman: This is to ignore the container components, which have a mapStateToProps, a mapDispatchToProps, and a call to Redux's connect HOC.

src/components/AssetsDropZone/container.jsx
```diff
@@ -0,0 +1,22 @@
+import { connect } from 'react-redux';
```
MichaelRoytman: This is just pulling the mapStateToProps, mapDispatchToProps, and connect out of AssetsDropZone/index.jsx into AssetsDropZone/container.jsx; there's no new code.

src/components/AssetsDropZone/index.jsx
```diff
@@ -1,18 +1,13 @@
...
 import Button from '@edx/paragon/src/Button';
 import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
-
-import { uploadAssets, uploadExceedMaxSize, uploadExceedMaxCount } from '../../data/actions/assets';
-
 import styles from './AssetsDropZone.scss';
 
-
-export class AssetsDropZone extends React.Component {
+class AssetsDropZone extends React.Component {
```
MichaelRoytman: This is just pulling the mapStateToProps, mapDispatchToProps, and connect out of AssetsDropZone/index.jsx into AssetsDropZone/container.jsx; there's no new code

src/components/AssetsFilters/container.jsx
```diff
@@ -0,0 +1,21 @@
+import { connect } from 'react-redux';
```
MichaelRoytman: This is just pulling the mapStateToProps, mapDispatchToProps, and connect out of AssetsFilters/index.jsx into AssetsFilters/container.jsx; there's no new code

src/components/AssetsFilters/index.jsx
```diff
@@ -32,22 +30,18 @@ AssetsFilters.propTypes = {
   assetsFilters: PropTypes.shape({
     assetTypes: PropTypes.object,
   }).isRequired,
+  courseDetails: PropTypes.shape({
```
MichaelRoytman: This is just pulling the mapStateToProps, mapDispatchToProps, and connect out of AssetsFilters/index.jsx into AssetsFilters/container.jsx; there's no new code

src/components/AssetsFilters/index.jsx
```diff
@@ -20,7 +18,7 @@ export const AssetsFilters = ({ assetsFilters, updateFilter, updatePage }) => (
             name={type.key}
             label={type.displayName}
             checked={assetsFilters.assetTypes[type.key]}
-            onChange={(checked) => { updatePage(0); updateFilter(type.key, checked); }}
+            onChange={(checked) => { updateFilter(type.key, checked, courseDetails); }}
```
MichaelRoytman: The action creator now handles updating the page, so the component no longer needs to do that.

src/components/AssetsPage/AssetsPage.test.jsx
```diff
@@ -1,25 +1,31 @@
...
-// import { types } from './pageTypes';
 
 const { shallow } = Enzyme;
 
 let wrapper;
 
 const defaultProps = {
   assetsList: [],
-  courseDetails: {},
+  courseDetails: {
```
MichaelRoytman: TODO: We're defining this in many tests. I'd like to pull it out into some kind test constants file.

src/components/AssetsPage/AssetsPage.test.jsx
```diff
@@ -319,72 +325,6 @@ describe('<AssetsPage />', () => {
       });
     });
   });
-  describe('getAssets', () => {
```
MichaelRoytman: I removed the mapDispatchToProps tests and the request state tests, since the action creators are now thunks (much harder to test) and request state has been removed from Redux.

src/components/AssetsPage/container.jsx
```diff
@@ -0,0 +1,25 @@
+import { connect } from 'react-redux';
```
MichaelRoytman: This is just pulling the mapStateToProps, mapDispatchToProps, and connect out of AssetsPage/index.jsx into AssestsPage/container.jsx; there's no new code

src/data/actions/assets.js
```diff
@@ -24,51 +16,120 @@ export const requestingAssets = () => ({
...
     dispatch(requestingAssets());
-    return clientApi.requestAssets(courseDetails.id, {
-      page: request.page,
-      assetTypes: request.assetTypes,
-      sort: request.sort,
-      direction: request.direction,
-    })
+
+    const state = getState();
+    const requestParameters = {
```
MichaelRoytman: Use the current state of the page to build request parameters to match, and then absorb the newly requested parameters.

src/data/reducers/assets.js
```diff
@@ -159,34 +156,6 @@ export const status = (state = {}, action) => {
   }
 };
 
-export const request = (state = requestInitial, action) => {
```
MichaelRoytman: 💀

src/data/actions/assets.js
```diff
@@ -24,51 +16,120 @@ export const requestingAssets = () => ({
...
-      .then((json) => {
-        const lastRequest = getState().request;
-        if (isSameResponse(request, lastRequest)) {
-          dispatch(requestAssetsSuccess(json));
-        }
-      })
-      .catch((error) => {
-        dispatch(requestAssetsFailure(error));
-      });
+      .then(json => (
```
MichaelRoytman: I believe that we no longer have potential for an infinite loop with the removal of the request state, since there is no longer a piece of state to which changes cause API calls that, in turn, cause changes to that very same state.

As far as race conditions go, I was reading a little about dealing with race conditions in React/Redux, and it sounds like our state organization was just [suboptimal](https://stackoverflow.com/questions/37738268/how-to-avoid-race-conditions-when-fetching-data-with-redux), and it should not be a problem. I'm not sure whether the response in that post maps perfectly to our state, though.

TODO: I have to give it a bit more thought. I feel like it's possible for, between two API calls A1 and A2, for A2 to return before A1. In that case, what happens to the state? Can I replicate this situation somehow?

package.json
```diff
@@ -91,7 +91,8 @@
       "/node_modules/",
       "/src/accessibilityIndex.jsx",
       "/src/index.jsx",
-      "/src/components/BackendStatusBanner/"
+      "/src/components/BackendStatusBanner/",
+      "<rootDir>/src/.*/container.jsx"
```
MichaelRoytman: Ignore coverage on container components (i.e. connected components, with mapStateToProps and mapDispatchToProps)

src/data/actions/assets.js
```diff
@@ -1,74 +1,184 @@
...
+  (dispatch, getState) => {
+    const currentPageState = getState().metadata.pagination;
+
+    const parameters = {
+      page,
+    };
+
+    return dispatch(getAssets(parameters, courseDetails)).then((responseAction) => {
+      if (requestFailed(responseAction)) {
+        // react-paginate does force the page state via the forcePage prop when
```
MichaelRoytman: Let me know if this is unclear.

src/data/reducers/assets.js
```diff
@@ -50,9 +54,9 @@ export const assets = (state = [], action) => {
   let assetsList;
   switch (action.type) {
     case assetActions.request.REQUEST_ASSETS_SUCCESS:
-      return action.data.assets;
-    case assetActions.delete.DELETE_ASSET_SUCCESS:
```
MichaelRoytman: Thoughts on removing this? This was doing a removal of the asset from the table when the page knew that the asset was successfully deleted. This was before I added in the following getAssets call to refresh the assets on the page. Leaving this in would do an optimistic removal of the asset from the table data before the getAssets call returns. This would cause a bit of page thrash if the asset was the last on the page (deleted -> no content view -> refreshed view, as opposed to deleted -> refreshed view). I could leave this in, but there would have to be more logic added to prevent the page thrash, and I didn't want to increase the scope any more.

src/components/AssetsTable/displayMessages.jsx
```diff
@@ -101,6 +101,11 @@ const messages = defineMessages({
     defaultMessage: 'Error uploading {assetName}. Try again.',
     description: 'Generic error message while uploading files',
   },
+  assetsTableGenericUpdateError: {
+    id: 'assetsTableGenericUpdateError',
+    defaultMessage: 'The action could not be completed. Refresh the page, and then try the action again.',
+    description: 'States that the action could not be completed and ask the user to refresh the page and try the action again',
```
efischer19: Nit: verb agreement, `States` and `asks`

src/components/AssetsTable/index.jsx
```diff
@@ -351,6 +349,22 @@ export class AssetsTable extends React.Component {
...
+        alertType = 'danger';
+        break;
+      case assetActions.paginate.PAGE_UPDATE_FAILURE:
+        alertDialog = genericUpdateError;
+        alertType = 'danger';
+        break;
+      case assetActions.sort.SORT_UPDATE_FAILURE:
+        alertDialog = genericUpdateError;
+        alertType = 'danger';
+        break;
```
efischer19: 👍 on this approach to error handling. These failures ought to be infrequent enough; the generic message will be fine

src/data/actions/assets.js
efischer19: Minor clarification: this comment should read `react-paginate does` **not** `force the page state`, correct?

src/data/reducers/assets.js
efischer19: I'm cool with removing it

src/data/actions/assets.js
MichaelRoytman: Oops! Yes! Thank you for the catch.

PR #111, opened 2018-02-05
EDUCATOR-2268 i18n.pre_validate

efischer19:
I'm not sure whether pursuing this line of attack is worthwhile or incredibly silly. But it seems to be viable, so that's something.

`transifex_input` was created via this command:
```
$(npm bin)/react-intl-gettext json2pot -d en src/data/i18n/default/src/components/ src/data/i18n/default/transifex_input.po
```

If we wanted, we could add this as an additional "preprocessing" step before bringing transifex into play. Transifex could then watch this file and spit out output in the form of `<lang code>/transifex_output.po`. I'm fairly confident that we could then wire up the auto-updating Jenkins jobs to run the complementary `po2json` command to insert usable json back.


efischer19:
This PR is review ready @edx/educator-dahlia, context available at https://openedx.atlassian.net/browse/EDUCATOR-2268

It does several things, most of which are fairly minor cleanup tasks:

- change `asset_name` to `assetName` in several places, this was causing warnings that were polluting Travis logs
- remove the "black magic" devstack commands from `Makefile`. I don't think anyone really uses these much anymore (correct me if you do!), and @thallada's recent work at https://github.com/edx/edx-platform/pull/17417 obviates them fairly well.
- moves Travis's package-lock check into `make package-lock.validate`, for clarity

The meat of the change is:

- add `react-intl-gettext`, which allows us to transform all our display strings into a single .po file
-add `i18n.*` make commands. `make i18n.docker` is the one devs will need to run, `make i18n.pre_validate` is what Travis runs to check that they did so.

Makefile
```diff
@@ -47,27 +47,23 @@ restart-detached:
...
-validate-devstack-folders:
-	export TEMP_FILE=$$(date +%s) && touch $$TEMP_FILE && docker exec -t edx.devstack.studio bash -c 'if [[ $$(ls /edx/src/sfe_parent/studio-frontend/$$TEMP_FILE) ]]; then echo "Folder structure looks good"; else echo "Your folders are FUBAR, talk to Eric (who should really come up with an automatic way of repairing this situation)"; exit 1; fi' && rm -rf $$TEMP_FILE
+i18n.docker: ## what devs should do from their host machines
+	docker exec -t dahlia.studio-frontend bash -c 'make i18n.extract && make i18n.preprocess'
+
+i18n.extract: ## move display strings from displayMessages.jsx to displayMessages.json
+	npm run-script i18n_extract
+
+i18n.preprocess: ## convert strings defined in displayMessages.json into .po file for Transifex
+	$$(npm bin)/react-intl-gettext json2pot -d en src/data/i18n/default/src/components/ src/data/i18n/default/transifex_input.po
```
fysheets: Nice!

.travis.yml
```diff
@@ -8,12 +8,13 @@ services:
  - docker
 
 before_script:
- - git diff --name-only --exit-code
  - docker build -t studio-frontend --no-cache .
 
 script:
   - npm run lint
   - npm run test
+  - make i18n.pre_validate
```
fysheets: Does this mean translations are run every time something goes to Travis?

efischer19: Yeah, it fails the build if you haven't run `make i18n.docker` before committing, similar to how `make package-lock.validate` will fail if you haven't committed package-lock updates.

Makefile
```diff
@@ -47,27 +47,23 @@ restart-detached:
 	make down
 	make up-detached
 
-devstack-install: validate-devstack-folders ## install local version of package into docker devstack for development
```
fysheets: Confirming this means that studio-frontend no longer needs to be nested in pre-existing folder structure, right?

package.json
```diff
@@ -64,6 +64,7 @@
     "postcss-loader": "^2.0.10",
     "postcss-prepend-selector": "^0.3.1",
     "react-dev-utils": "^4.0.0",
+    "react-intl-gettext": "^5.1.1",
     "react-test-renderer": "^16.1.0",
```
fysheets: Not related to this, but this still needs a version bump, yeah?

.travis.yml
```diff
@@ -8,12 +8,13 @@ services:
...
+  - make package-lock.validate
```
thallada: The downside to this is that you won't know until after the full ~5 minutes that this is why the build failed. If it is in before_script then it will be run before anything else and will immediately fail the build without running any tests.

Makefile
efischer19: yes. With tyler's incoming changes, you can have sfe anywhere relative to devstack and they'll be able to talk

package.json
efischer19: does "this" refer to package.json in general? We do need to add that still, I was going to coordinate with @thallada since I think https://github.com/edx/studio-frontend/pull/112 is landing shortly too

package-lock.json
```diff
@@ -4089,6 +4090,910 @@
       "resolved": "https://registry.npmjs.org/fs.realpath/-/fs.realpath-1.0.0.tgz",
       "integrity": "sha1-FQStJSMVjKpA20onh8sBQRmU6k8="
     },
+    "fsevents": {
```
MichaelRoytman: It looks like the fsevents package, et. al. are being added again. I was able to get rid of it by `rm -rf node-modules`, bringing the container down, bringing the container up, and letting the container do the npm install, and that seemed to resolve it for me.

.travis.yml
efischer19: That's interesting - I had moved it here because I noticed that even if these commands fail, the other commands around them will continue, as seen here: https://travis-ci.org/edx/studio-frontend/builds/338187324

This was helpful in the PR specifically, as I had all of linting, test, and i18n.pre_validate errors at various times and I was able to work on them in parallel. Should we prefer fast, sequential failures or longer, parallel full test suite runs?

Also worth pointing out - in the link above, the full run took 2:35, 1:14 of which was spent running `docker build -t studio-frontend --no-cache .`. Do we really need to continue using docker inside Travis? None of the other test commands make use of it.

package-lock.json
thallada: Also make sure you have NPM 5.6.0. It was supposedly fixed in that version, but there is still [an issue open saying that it still doesn't work sometimes](https://github.com/npm/npm/issues/17722)...

.travis.yml
thallada: Yeah that's fair, I guess it's better to have the full test results.

If the tests work without the docker container, then I think we should do away with that line. I can't think of any reason we would need it. We could always add it back if needed.

PR #112, opened 2018-02-06
Use built files in platform & allow development inside devstack Studio shell

thallada:
[See the platform PR which will use these changes.](https://github.com/edx/edx-platform/pull/17417) It might make more sense with the context of both PRs.

1. Connect our `dahlia.studio-frontend` docker container to the same network as the devstack containers so that they can talk to each other.
    * This will allow us to develop on studio-frontend inside of devstack's Studio!
2. Allow other domains (devstack studio) to access the studio-frontend webpack-dev-server
    * Also for developing inside devstack studio
3. Use CommonChunksPlugin in prod
    * This is so that each studio-frontend component can share the same (cached) vendor files (react, paragon, bootstrap, etc.)
4. Styling tweaks so that the font-sizes match between bare studio-frontend and studio-frontend inside devstack studio.
5. Got rid of the linux-specific docker-compose stuff. I'm using linux and everything worked fine with the normal docker-compose file.
    * I'm not really sure if this will make Travis break or not though...

I want to get this PR merged first as a tagged release so that I can install the version in the platform PR and build a sandbox off it to test.

Edit: I've now pulled in the "step 2" of CSS fixes which involves the `all: initial` hack. I was having some troubles integrating into Studio because it has `.modal` styles which clobber our bootstrap `.modal` styles.


config/webpack.dev.config.js
```diff
@@ -7,28 +7,10 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
 const apiEndpoints = require('../src/data/api/endpoints.js');
 const commonConfig = require('./webpack.common.config.js');
 
-let targetUrl = 'localhost';
-if (!process.env.RUNNING_ON_LINUX) {
-  targetUrl = `docker.for.mac.${targetUrl}`;
-}
-targetUrl = `http://${targetUrl}:18010`;
+const targetUrl = 'http://edx.devstack.studio:18010';
```
efischer19: nice

thallada:
Note for standup tomorrow: someone on a mac should pull this branch and recreate their docker container to test that everything still works okay.

README.md
```diff
@@ -57,8 +73,9 @@ Currently, the process for releasing a new version of our package is as follows:
 2. Merge your pull request.
 3. Publish a [GitHub release](https://github.com/edx/studio-frontend/releases). Make sure to prefix the version number with `v`, as in `v2.3.4`.
 3. `git checkout master` and `git pull`. Ensure your current directory is cleaned, with no outstanding commits or files. As an extra precaution, you can `rm -rf node node_modules` and `npm install` prior to publishing.
-4. Be a member of the correct edX and npm orgs, and be logged in. All of @edx/educator-dahlia should be set up, and others shouldn't need to be publishing this package.
-5. Run `npm publish`.
+4. Make sure that build production files in /dist will be included in the release by running `npm run build`.
```
thallada: I added this because I noticed that the npm prepublish step failed for me inside the studio-frontend container. I think it is because it was running as root and NPM doesn't like that.

How much trouble would it be to log into the studio-frontend container not as root (e.g. devstack uses "edxapp" user)?

config/webpack.dev.config.js
```diff
@@ -7,28 +7,10 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
...
 
 module.exports = Merge.smart(commonConfig, {
   devtool: 'cheap-module-eval-source-map',
-  entry: {
-    assets: [
-      // enable react's custom hot dev client so we get errors reported
-      // in the browser
-      require.resolve('react-dev-utils/webpackHotDevClient'),
```
thallada: The stock webpack-dev-server can do the error overlay too, so I removed this in favor of a simpler webpack config and less dependencies.

config/webpack.dev.config.js
```diff
@@ -62,7 +44,8 @@ module.exports = Merge.smart(commonConfig, {
               plugins: () => [
                 /* eslint-disable global-require */
                 require('autoprefixer'),
-                require('postcss-prepend-selector')({ selector: '.SFE ' }),
+                require('postcss-initial')(),
+                require('postcss-prepend-selector')({ selector: '#root.SFE ' }),
```
thallada: I had to make this selector more specific because there are some legacy Studio selectors that were still getting past my `all: initial` wall because they were overly specific. It's possible there are even more specific selectors that this doesn't override, but there at least aren't any on the Files & Uploads page.

config/webpack.dev.config.js
```diff
@@ -102,6 +85,10 @@ module.exports = Merge.smart(commonConfig, {
   devServer: {
     host: '0.0.0.0',
     port: 18011,
+    headers: {
+      'Access-Control-Allow-Origin': '*',
```
thallada: This allows Studio devstack, which is running on another "domain": `localhost:18010`, to access the webpack-dev-server on this docker container (`dahlia.studio-frontend`). Theoretically this would allow any other website on the internet to access your SFE assets, but :man_shrugging: don't put anything secret there.

config/webpack.prod.config.js
```diff
@@ -61,13 +68,25 @@ module.exports = Merge.smart(commonConfig, {
...
   plugins: [
-    new ExtractTextPlugin('studio-frontend.min.css'),
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor',
+      minChunks: module => module.context && module.context.includes('node_modules'),
+    }),
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'manifest',
+      minChunks: Infinity,
+    }),
```
thallada: This webpack incantation is explained here: https://webpack.js.org/plugins/commons-chunk-plugin/#combining-implicit-common-vendor-chunks-and-manifest-file

public/index.html
```diff
@@ -1,6 +1,12 @@
 <!doctype html>
 <html>
-  <head></head>
+  <head>
+    <style>
+      html {
+        font-size: 62.5%;
+      }
+    </style>
+  </head>
```
thallada: Studio has this style, so I'm replicating it here. Without this style, the fonts in bare studio-frontend would be ginormous because all of the rem units are assuming that this is the base font size.

src/_edx-bootstrap.scss
```diff
@@ -2,8 +2,10 @@
  * (those should go into SFE.scss). Because this file does not define styles, it is okay to import into every
  * individual component's SCSS file. */
 
+$base-rem-size: 0.625;
 @import '~@edx/edx-bootstrap/sass/edx/theme';
 @import '~bootstrap/scss/variables';
 @import '~bootstrap/scss/mixins';
 @import '~bootstrap/scss/utilities/spacing';
 @import '~bootstrap/scss/utilities/screenreaders';
+@import 'var-reset';
```
thallada: Unfortunately we can't use the edx-boostrap legacy reset scss here because it also includes a lot of CSS definitions (the bootstrap reboot sass) that I don't want to include on every single component. This is just the variable reset portion of that file.

PR #129, opened 2018-02-21
add data-identifier data attributes to HTML elements for easier selec…

MichaelRoytman:
…tion

### DIFF
--- a/src/components/AssetsTable/AssetsTable.test.jsx
+++ b/src/components/AssetsTable/AssetsTable.test.jsx
@@ -173,7 +173,9 @@ describe('<AssetsTable />', () => {
       const baseUrl = defaultProps.courseDetails.base_url;
 
       rows.forEach((row, index) => {
-        expect(row.containsMatchingElement(<td>{defaultProps.assetsList[index].display_name}</td>))
+        expect(row.containsMatchingElement(
+          <td><span data-identifier="asset-file-name">{defaultProps.assetsList[index].display_name}</span></td>,
+        ))
           .toEqual(true);
 
         const rowThumbnail = defaultProps.assetsList[index].thumbnail;
--- a/src/components/AssetsTable/index.jsx
+++ b/src/components/AssetsTable/index.jsx
@@ -131,13 +131,13 @@ export default class AssetsTable extends React.Component {
     if (thumbnail) {
       return (
         <WrappedMessage message={messages.assetsTableNoDescription} >
-          { displayText => (<img src={`${baseUrl}${thumbnail}`} alt={displayText} />) }
+          { displayText => (<img src={`${baseUrl}${thumbnail}`} alt={displayText} data-identifier="asset-image-thumbnail" />) }
         </WrappedMessage>
       );
     }
     return (
       <WrappedMessage message={messages.assetsTableNoPreview} >
-        { displayText => <div className={styles['no-image-preview']}>{displayText}</div> }
+        { displayText => <div className={styles['no-image-preview']} data-identifier="asset-image-thumbnail">{displayText}</div> }
       </WrappedMessage>
     );
   }
@@ -161,6 +161,7 @@ export default class AssetsTable extends React.Component {
             data-asset-id={asset.id}
             aria-label={displayText}
             onClick={this.onLockClick}
+            data-identifier="asset-lock-button"
           />)
         }
       </WrappedMessage>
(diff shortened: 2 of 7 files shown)


src/components/AssetsTable/index.jsx
```diff
@@ -564,11 +583,7 @@ export default class AssetsTable extends React.Component {
       />
     );
 
-    return (
-      <React.Fragment>
-        {statusAlert}
-      </React.Fragment>
-    );
+    return statusAlert;
```
MichaelRoytman: I don't think we need the React.Fragment here, since it's a single element, but please correct me if I'm wrong.

src/components/CopyButton/index.jsx
```diff
@@ -69,6 +69,7 @@ export default class CopyButton extends React.Component {
         onBlur={this.onBlur}
         onMouseUp={this.onMouseUp}
         onMouseDown={this.onMouseDown}
+        data-identifier={this.props['data-identifier']}
```
MichaelRoytman: I had to add this data-identifier prop to CopyButton to add it to the underlying HTML button.

src/components/AssetsTable/index.jsx
```diff
@@ -526,6 +544,7 @@ export default class AssetsTable extends React.Component {
         }
       </WrappedMessage>
     );
+
     return (
       <React.Fragment>
```
thallada: This doesn't need to be a `React.Fragment` too, right?

src/components/CopyButton/index.jsx
```diff
@@ -81,10 +82,12 @@ CopyButton.propTypes = {
...
   onCopyButtonClick: PropTypes.func,
   className: PropTypes.arrayOf(PropTypes.string),
+  'data-identifier': PropTypes.string,
 };
 
 CopyButton.defaultProps = {
   onCopyButtonClick: () => {},
   ariaLabel: 'Copy',
   className: [],
+  'data-identifier': 'copy-button',
```
thallada: Are we expecting `data-identifier` to be a unique id? It might be better to leave this `undefined` and only add the data attribute to the element when it is defined. Otherwise we might use CopyButton for something else in the future, forget to specify `data-identifier` because it's not required, and then end up with multiple elements with the same `data-identifier="copy-button"`.

src/components/AssetsTable/AssetsTable.test.jsx
```diff
@@ -173,7 +173,9 @@ describe('<AssetsTable />', () => {
       const baseUrl = defaultProps.courseDetails.base_url;
 
       rows.forEach((row, index) => {
-        expect(row.containsMatchingElement(<td>{defaultProps.assetsList[index].display_name}</td>))
+        expect(row.containsMatchingElement(
+          <td><span>{defaultProps.assetsList[index].display_name}</span></td>,
```
fysheets: Does the new span here need to have a data-identifier attribute on it?

src/components/AssetsTable/index.jsx
```diff
@@ -161,6 +161,7 @@ export default class AssetsTable extends React.Component {
             data-asset-id={asset.id}
             aria-label={displayText}
             onClick={this.onLockClick}
+            data-identifier="asset-lock-button"
```
fysheets: Is it possible for this to be lock/unlock depending on which class it is?

fysheets: Actually nvm, it will be possible to fetch lock/unlock with the regular CSS selector :) disregard.

src/components/AssetsTable/index.jsx
```diff
@@ -429,6 +434,18 @@ export default class AssetsTable extends React.Component {
         currentAsset.external_url,
       );
 
+      currentAsset.display_name = (
+        <span data-identifier="asset-file-name">{currentAsset.display_name}</span>
+      );
+
+      currentAsset.content_type = (
+        <span data-identifier="asset-content-type">{currentAsset.display_name}</span>
```
fysheets: Is this also supposed to be `currentAsset.display_name`?

src/components/AssetsTable/index.jsx
MichaelRoytman: It has to be, since the two WrappedMessage elements are adjacent and need to be wrapped in something.

src/components/CopyButton/index.jsx
MichaelRoytman: Ideally, yes, it would be unique. That's a good idea. I'll change it to be undefined.

src/components/AssetsTable/index.jsx
MichaelRoytman: It's supposed to be content_type. I copy-pasted. Thanks for catching that.

PR #132, opened 2018-02-22
Wrapped search behind feature flag

dsjen:
This PR is against @thallada's feature branch. I've wrapped the search component within a conditional, added tests, and the feature flag/switch setting. If you think that the interface for the setting looks correct, I can go ahead with the relevant platform PR to pass the setting over.

Please review @edx/educator-dahlia

### DIFF
--- a/src/components/AssetsPage/index.jsx
+++ b/src/components/AssetsPage/index.jsx
@@ -151,11 +151,13 @@ export default class AssetsPage extends React.Component {
     return (
       <div className={styles.assets}>
         <div className={edxBootstrap.container}>
-          <div className={edxBootstrap.row}>
-            <div className={`${edxBootstrap['col-md-3']} ${edxBootstrap['offset-md-9']}`}>
-              <WrappedAssetsSearch />
+          {this.props.searchSettings.enabled &&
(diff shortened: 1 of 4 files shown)


src/components/AssetsPage/index.jsx
```diff
@@ -151,11 +151,13 @@ export default class AssetsPage extends React.Component {
     return (
       <div className={styles.assets}>
         <div className={edxBootstrap.container}>
-          <div className={edxBootstrap.row}>
-            <div className={`${edxBootstrap['col-md-3']} ${edxBootstrap['offset-md-9']}`}>
-              <WrappedAssetsSearch />
+          {this.props.searchSettings.enabled &&
```
dsjen: I modeled this conditional rendering from a [reactjs example](https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator).

thallada:
Looks like the tests are failing on translations stuff. I ran `make docker.i18n` on my checkout which might fix that, but I haven't committed it yet. Doesn't matter if the tests are failing anyways if this is merging into `thallada/search`.

dsjen:
Sounds good, @thallada!

PR #134, opened 2018-02-24
EDUCATOR-2434 - hiding image previews with a button

MichaelRoytman:
I just wanted to quickly see how easy it would be to hide the image previews. I didn't think too much about the button styling.

Using this code I found online to get the length of the page:

```
var body = document.body,
    html = document.documentElement;

var height = Math.max( body.scrollHeight, body.offsetHeight, 
                       html.clientHeight, html.scrollHeight, html.offsetHeight );
```

from [here](https://stackoverflow.com/questions/1145850/how-to-get-height-of-entire-document-with-javascript),  I found that, in the demo course with the image filter enabled, the length of the page with image previews is 4,129 pixels, and the length of the page without image previews is 2,722, which is around a 33% decrease.

For a page that is a mixture of asset types (page 22 in the demo course with no filtering enabled), the length of the page with image previews is 6,541, and the length of the page without image previews is 6,482. That's not nearly as significant, but I believe it's because some of the longer file names are forcing the copy buttons to stack, which increases the size of each row. There is also  lot of wasted padding space between the file name and the content type, so hiding the image preview does not has as dramatic an effect. It might be interesting to look at ways to utilize the space more efficiently.

### DIFF
--- a/src/components/AssetsImagePreviewFilter/displayMessages.jsx
+++ b/src/components/AssetsImagePreviewFilter/displayMessages.jsx
@@ -1,0 +1,11 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  assetsImagePreviewFilterLabel: {
+    id: 'assetsImagePreviewFilterLabel',
+    defaultMessage: 'Hide File Preview',
+    description: 'Label checkbox to hide image previews in the assets table.',
+  },
+});
+
+export default messages;
--- a/src/components/AssetsImagePreviewFilter/index.jsx
+++ b/src/components/AssetsImagePreviewFilter/index.jsx
@@ -1,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CheckBox } from '@edx/paragon';
+
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
+import messages from './displayMessages';
+
+const AssetsImagePreviewFilter = ({ isImagePreviewEnabled, updateImagePreview }) => (
+  <CheckBox
+    id="imagePreviewCheckbox"
+    name="imagePreviewCheckbox"
+    label={<WrappedMessage message={messages.assetsImagePreviewFilterLabel} />}
+    checked={!isImagePreviewEnabled}
+    onChange={(checked) => { updateImagePreview(!checked); }}
+  />
+);
+
+AssetsImagePreviewFilter.propTypes = {
+  isImagePreviewEnabled: PropTypes.bool.isRequired,
+  updateImagePreview: PropTypes.func.isRequired,
+};
+
+export default AssetsImagePreviewFilter;
--- a/src/components/AssetsPage/index.jsx
+++ b/src/components/AssetsPage/index.jsx
@@ -4,10 +4,10 @@ import PropTypes from 'prop-types';
 import { assetActions } from '../../data/constants/actionTypes';
 import { hasSearchOrFilterApplied } from '../../utils/getAssetsFilters';
 import edxBootstrap from '../../SFE.scss';
-import styles from './AssetsPage.scss';
 import WrappedAssetsDropZone from '../AssetsDropZone/container';
 import WrappedAssetsTable from '../AssetsTable/container';
 import WrappedAssetsFilters from '../AssetsFilters/container';
+import WrappedAssetsImagePreviewFilter from '../AssetsImagePreviewFilter/container';
 import WrappedPagination from '../Pagination/container';
 import WrappedAssetsSearch from '../AssetsSearch/container';
 import WrappedAssetsStatusAlert from '../AssetsStatusAlert/container';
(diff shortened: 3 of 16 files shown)


src/components/AssetsImagePreviewFilter/displayMessages.jsx
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  assetsImagePreviewFilterLabel: {
+    id: 'assetsImagePreviewFilterLabel',
+    defaultMessage: 'Hide File Preview',
```
MichaelRoytman: TODO: Check in with Sylvia.

src/components/AssetsImagePreviewFilter/index.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CheckBox } from '@edx/paragon';
+
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
+import messages from './displayMessages';
+
+const AssetsImagePreviewFilter = ({ updateImagePreview, isImagePreviewEnabled }) => (
+  <CheckBox
+    // key={type.key} // I don't know if we need this. It's only for lists of elements, right?
```
MichaelRoytman: I'm not sure what these three values should be. I copied the CheckBox from the AssetsFilters. I don't think we need key, but I'm not sure about id and name.

src/components/AssetsPage/index.jsx
```diff
@@ -85,6 +86,8 @@ export default class AssetsPage extends React.Component {
     <React.Fragment>
       <div className={edxBootstrap.col}>
         { this.renderAssetsDropZone() }
+        <hr />
```
MichaelRoytman: This adds the line break. Right now, the break is gray. I know the mock-up had a black line. There's [this variable](https://github.com/edx/edx-bootstrap/blob/f0a6237174de7dd65815b07c73da127f082faa4d/sass/edx/_variables.scss#L276) commented out in edx-bootstrap. Should we uncomment it?

src/components/AssetsImagePreviewFilter/index.jsx
```diff
@@ -0,0 +1,24 @@
...
+    // id={type.key} // what should this be?
+    name="imagePreviewEnabled" // what should this be?
+    label={<WrappedMessage message={messages.assetsImagePreviewFilterLabel} />}
+    checked={!isImagePreviewEnabled}
```
MichaelRoytman: I'm not sure how I feel about this. The checkbox "check" state and the actual meaning of that state are in conflict. If you check the checkbox, it disables the image preview, which necessitates the `!`. So, if the checkbox checked state is `true`, isImagePreviewEnabled is `false`. I wonder whether the checkbox label should be "Show Image Previews", and it should render checked by default. That could make the code cleaner, since a checkbox state of `true` would correspond to `true` for isImagePreviewEnabled. Just a thought. I don't think this code is terribly atrocious; it's just not super intuitive.

src/components/AssetsImagePreviewFilter/index.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { CheckBox } from '@edx/paragon';
+
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
+import messages from './displayMessages';
+
+const AssetsImagePreviewFilter = ({ updateImagePreview, isImagePreviewEnabled }) => (
```
MichaelRoytman: Should this be "file preview"? I guess the preview is displayed for all files, making it a file preview, but it's also intended only for image files, so... Should the name match the check box label?

src/data/reducers/assets.js
```diff
@@ -181,6 +184,17 @@ export const status = (state = {}, action) => {
   }
 };
 
+export const imagePreview = (state = imagePreviewInitial, action) => {
```
MichaelRoytman: Because I made the filter a separate element, it is rendered by the AssetsPage. The state that the AssetsImagePreviewFilter component is responsible for is also needed by the AssetsTable in order to show/hide the image preview column. In order for them to both have that piece of state, I could either lift the state up into the AssetsPage or move it into Redux. I decided to go with Redux, since I thought it would be a little more intuitive, but we could just lift the state up. AssetsPage could pass a callback into the AssetsImagePreviewFilter and then pass a isImagePreviewEnabled prop to the AssetsTable. Any opinions?

src/components/AssetsPage/index.jsx
fysheets: There is a bootstrap class name that can add the line -- I believe it's the `page-header` class name https://getbootstrap.com/docs/3.3/components/#page-header

If that doesn't work I can dig in deeper.

src/components/AssetsImagePreviewFilter/displayMessages.jsx
MichaelRoytman: Sylvia [said](https://openedx.atlassian.net/browse/EDUCATOR-2434?focusedCommentId=311150&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-311150) this is good to go.

PR #136, opened 2018-02-26
locale data scripts are no longer passed in DOM

efischer19:
This is the result of Friday's flailing about - instead of passing a single locale data script over in the DOM, we'll let all 5 of them be bundled into SFE together and choose which too use dynamically at runtime, using server-specified locale.

The reason we are electing to do this is that it turned out to be waaaaay to difficult to only pass a single locale's data script while also avoiding a client-server roundtrip. There are only 5 currently supported languages, and their data scripts are fairly small, so let's just do the simple thing here instead of over-optimizing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,6 +8,7 @@
     "i18n_extract": "BABEL_ENV=i18n $(npm bin)/babel src --quiet > /dev/null",
     "i18n_cleanJson": "node scripts/cleanI18nJson.js",
     "lint": "eslint --ext .js --ext .jsx .",
+    "lint_fix": "eslint --fix --ext .js --ext .jsx .",
     "prepublishOnly": "npm run build",
     "start": "NODE_ENV=development BABEL_ENV=development $(npm bin)/webpack-dev-server --config=config/webpack.dev.config.js --hot --inline --progress",
     "test": "jest --coverage"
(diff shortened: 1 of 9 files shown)


src/data/i18n/locales/json/currentlySupportedLangs.jsx
```diff
@@ -14,3 +14,12 @@ import './zh_CN.json';
 //
 // Further, I hit the read replica in order to confirm the live values, which are currently:
 //    en, fr, zh-cn, es-419, ar
+
+const localeDataMap = {
+  'ar': arData,
+  'es-419': esData,
+  'fr': frData,
+  'zh-cn': zhData,
```
efischer19: I added the `quote-props` rule with `consistent-as-needed` to our `eslintrc` because of this block - otherwise it would have me leave `ar` and `fr` unquoted, but require quote around `es-419` and `zh-cn`, which seems silly and odd.

package.json
```diff
@@ -8,6 +8,7 @@
     "i18n_extract": "BABEL_ENV=i18n $(npm bin)/babel src --quiet > /dev/null",
     "i18n_cleanJson": "node scripts/cleanI18nJson.js",
     "lint": "eslint --ext .js --ext .jsx .",
+    "lint_fix": "eslint --fix --ext .js --ext .jsx .",
```
efischer19: I also added this command and ran it, that's why a few of these seemingly unrelated files are in the diff.

PR #137, opened 2018-02-26
change url to portable_url for Studio URL copy button

MichaelRoytman:


dsjen:
Thanks for all the work in triaging and getting this out, team!

PR #138, opened 2018-02-26
EDUCATOR-1227 Assets Search UI

thallada:
- Adds AssetsSearch
- Adds AssetsResultCount
- Adds AssetsClearFiltersButton
- Restructures AssetsPage DOM to include new components
- Translates the strings in AssetsPage
- Adds new SEARCH actions/reducers
- Updates the clearFilters action so that it also clears the current search
- Conditionally show the new UI with studioContext setting (thanks @dsjen!)

I decided to pull the status alert refactor out into [a separate PR](https://github.com/edx/studio-frontend/compare/thallada/search...thallada/refactor-status-alert) since it was getting pretty gnarly.

~~I'll update with a comment once a sandbox for this PR is created.~~ I need to merge this PR and then make a separate one in platform and then I will be able to make a sandbox.

@edx/educator-dahlia


MichaelRoytman:
If you want, couldn't you install studio-frontend into edx-platform using a Git hash just to get a sandbox up?

src/components/AssetsClearFiltersButton/index.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Button } from '@edx/paragon';
+
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
+import messages from './displayMessages';
+
+
+const AssetsClearFiltersButton = ({ clearFilters, courseDetails }) => (
```
MichaelRoytman: Nit: If this button clears both the filters and the search now, I wonder if we can find a better name for it. I can't really think of a good one, so if no one else is opposed, it's fine. Maybe, `AssetsShowAllFilesButton`? Although, like Farhanah mentioned, I could see that being confusing too.

src/components/AssetsSearch/index.jsx
```diff
@@ -0,0 +1,89 @@
...
+    this.props.updateSearch(this.state.value, this.props.courseDetails);
+    event.preventDefault(); // prevent the page from navigating to the form action URL
+  }
+
+  handleChange(value) {
+    this.setState({ value });
+  }
+
+  render() {
+    // TODO: InputText creates it's own form-group div. Nesting them is not semantic.
```
dsjen: Nit: spelling on "it's" :)

src/components/AssetsSearch/AssetsSearch.scss
```diff
@@ -0,0 +1,9 @@
+.form-inline label {
+  margin-right: 10px;
+  font-weight: 600;
```
dsjen: Can we use [$font-weight-bold]($font-weight-bold) here? Doesn't everyone love theming?

src/components/AssetsSearch/AssetsSearch.scss
```diff
@@ -0,0 +1,9 @@
+.form-inline label {
+  margin-right: 10px;
```
dsjen: I think this could be `$spacer` just in case someone wants to theme spacing, sort of like what we do in https://github.com/edx/studio-frontend/blob/master/src/components/AccessibilityPolicyForm/AccessibilityPolicyForm.scss#L4

src/components/AssetsSearch/AssetsSearch.scss
dsjen: I guess alternatively one could extend the `font-weight-bold` css class too--seems like overkill perhaps though.

src/components/AssetsResultsCount/AssetsResultsCount.scss
```diff
@@ -0,0 +1,3 @@
+.result-count-wrapper {
+  padding: 10px;
```
dsjen: Maybe use [$spacer](https://github.com/edx/edx-bootstrap/blob/master/sass/open-edx/_variables.scss#L120)?

src/utils/getAssetsFilters.jsx
```diff
@@ -31,3 +31,7 @@ export const getFilterStateObjectFromArray = (enabledFilters) => {
 
   return filterState;
 };
+
+export const hasSearchOrFilterApplied = (filters, search) => (
+  Boolean(hasSelectedFilters(filters) || (search && search.length > 0))
```
dsjen: Out of curiosity, why cast it to a `Boolean`?

src/utils/getAssetsFilters.test.jsx
```diff
@@ -77,4 +77,37 @@ describe('getAssetsFilters Utility Functions', () => {
       expect(filterStateObject).toEqual(expectedfilterStateObject);
     });
   });
+  describe('hasSearchOrFilterApplied', () => {
```
dsjen: Nice testing!

src/components/AssetsPage/index.jsx
```diff
@@ -88,17 +88,40 @@ export default class AssetsPage extends React.Component {
...
-        <WrappedAssetsTable />
-        <WrappedPagination />
+        {this.props.searchSettings.enabled &&
+          <React.Fragment>
+            <div className={edxBootstrap.row}>
+              <div className={`${edxBootstrap['col-md-8']}`}>
+                <WrappedAssetsResultsCount />
+              </div>
+              <div className={`${edxBootstrap['col-md-4']} ${edxBootstrap['text-right']}`}>
+                {hasSearchOrFilterApplied(this.props.filtersMetaData.assetTypes,
```
dsjen: Thanks for covering this use case!

src/components/AssetsSearch/index.jsx
thallada: Ugh, I always get that wrong.

src/utils/getAssetsFilters.jsx
thallada: If `search` is set to something falsey, the expression will just return the value. Wrapping it in this cast will make sure it only ever returns `true` or `false`.

src/components/AssetsSearch/AssetsSearch.scss
thallada: So `$spacer` ends up being 16px because of the weird `base-rem-size` Studio has. I could do `1rem` instead for 10px and `0.8rem` for 8px. It would still be kinda configurable if they changed the `base-rem-size` (everything would scale with that).

src/components/AssetsSearch/AssetsSearch.scss
thallada: I tried that but it is set to 700 and I think it looks too "heavy". What do you think?
screenshot-20180226174240-472x84

vs. what it is set to now (600):
screenshot-20180226174820-461x89

thallada:
@MichaelRoytman I think I finally found a way to do what you mentioned. The key was changing our `prepublishOnly` script that runs "npm run build" to `prepare` instead, which will run when studio-frontend is installed as a github URL, but not when installed from the NPM registry.

> `npm install <git remote url>`:
...
If the package being installed contains a prepare script, its dependencies and devDependencies will be installed, and the prepare script will be run, before the package is packaged and installed.
-- https://docs.npmjs.com/cli/install

So, when the sandbox runs `npm install` it will pull down this branch and then run the prepare script which will run the webpack build and emit the /dist files that are needed for the later static asset steps. I'm starting a sandbox build now.

Unrelated to all of that, but I noticed that `npm run build` was failing on this branch (seemed to be caused by paragon 2.1.0) with a weird uglifyjs error, but I fixed it by upgrading the uglifyjs plugin :man_shrugging: . The version included with webpack 3 is pretty old.

A side-effect of changing to the `prepare` script is that Travis will now run `npm run build`. It will slow our builds down by like 35 seconds but I think it is worth it so we catch errors with the production build early.

src/components/AssetsSearch/AssetsSearch.scss
dsjen: Yeah, I see what you mean! @abbyluke, do you have an opinion on this? Is the [edx-bootstrap bold](https://github.com/edx/edx-bootstrap/blob/f0a6237174de7dd65815b07c73da127f082faa4d/sass/edx/_variables.scss#L237) in this case too bold?

src/utils/getAssetsFilters.jsx
dsjen: Gotcha. Thanks!

efischer19:
A thumb on the comment feels insufficient here - great find @thallada!

src/components/AssetsResultsCount/displayMessages.jsx
```diff
@@ -0,0 +1,16 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  assetsResultsCountFiltered: {
+    id: 'assetsResultsCountFiltered',
+    defaultMessage: 'Showing {start}-{end} out of {total} possible matches.',
```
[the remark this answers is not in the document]
efischer19: I'm curious to see how this gets represented as it goes up to and back from Transifex.

src/components/AssetsResultsCount/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+import { hasSearchOrFilterApplied } from '../../utils/getAssetsFilters';
+import styles from './AssetsResultsCount.scss';
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
+import messages from './displayMessages';
+
+const renderCount = count => (
+  <span className={`${edxBootstrap['font-weight-bold']}`}>
+    <FormattedNumber value={count} />
+  </span>
+);
```
efischer19: This function is super neat

🆒

src/components/AssetsResultsCount/index.jsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  return (
+    <div className={styles['result-count-wrapper']}>
+      <WrappedMessage
+        message={message}
+        values={{
+          start: renderCount(paginationMetaData.start + 1),
+          end: renderCount(paginationMetaData.end),
+          total: renderCount(paginationMetaData.totalCount),
```
efischer19: I'm not sure if using `renderCount` like this will work correctly with the pluralization I mentioned above, I think that needs a plain number. Surely there's something for this scenario...

LMK when you get to this portion, I'll help get it ironed out if it's tricky. I'm also okay with ignoring the pluralization work in the short term (still to be done later) if it's going to delay this large PR from landing.

src/data/i18n/default/src/components/AssetsClearFiltersButton/displayMessages.json
```diff
@@ -0,0 +1,7 @@
+[
+  {
+    "id": "assetsClearFiltersButtonLabel",
+    "description": "Label for a button that clears all filters applied to a table.",
+    "defaultMessage": "Clear all filters"
+  }
+]
```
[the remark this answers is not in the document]
efischer19: These are tool-generated; all `<component>/displayMessages.json files under `src/data/i18n/default` will be this way.

src/components/AssetsPage/container.jsx
```diff
@@ -10,10 +10,11 @@ const mapStateToProps = state => ({
   uploadSettings: state.studioDetails.upload_settings,
   status: state.metadata.status,
   filtersMetaData: state.metadata.filters,
+  searchMetaData: state.metadata.search,
```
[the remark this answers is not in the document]
thallada: I agree. I like `Metadata` over `MetaData`. I can just do a find and replace on that.

src/components/AssetsPage/index.jsx
```diff
@@ -88,17 +88,40 @@ export default class AssetsPage extends React.Component {
...
       </div>
     </React.Fragment>
   );
 
   renderNoAssetsBody = () => (
-    <React.Fragment>
-      <h3>0 files in your course</h3>
-      <h4>Enhance your course content by uploading files such as images and documents.</h4>
-    </React.Fragment>
+    <div>
```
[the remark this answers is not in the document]
thallada: I think I might have done that on your suggestion but forgot to change it back when I realized that wasn't the issue.

src/components/AssetsClearFiltersButton/index.jsx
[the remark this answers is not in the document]
thallada: Yeah I'm not sure where to go with this. To me a search is something you use to filter, but we have also distinguished filters as something different than search in the code. I like calling it a "clear" because that's the action it performs, but what is it clearing? `AssetsClearFiltersAndSearchButton`, `AssetsClearControlsButton`, `AssetsClearQueryButton`...?

src/components/AssetsClearFiltersButton/displayMessages.jsx
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  assetsClearFiltersButtonLabel: {
+    id: 'assetsClearFiltersButtonLabel',
+    defaultMessage: 'Clear all filters',
```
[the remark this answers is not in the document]
thallada: I think Sylvia had concerns over the wording when it was always displayed. Maybe it's not an issue when it's hidden while no search or filters are applied. Whatever we decide for the variable name I think we should use here too.

src/data/i18n/default/src/components/AssetsPage/displayMessages.json
```diff
@@ -0,0 +1,27 @@
+[
+  {
+    "id": "assetsPageNoResultsNumFiles",
```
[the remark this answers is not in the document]
thallada: Yeah, I should be more consistent. `Count` works for me :+1:

src/components/AssetsResultsCount/index.jsx
thallada: Maybe [`<FormattedPlural>`](https://github.com/yahoo/react-intl/wiki/Components#formattedplural) could do it? It looks like it allows embedding the text inside elements (just like `<FormattedMessage>`). Since this means we will have to make separate displayMessages for "file"/"total files" and "possible match"/"possible matches" strings, we would need to make a `<WrappedFormattedPlural>` so each plural string is wrapped in the proper lang.

I still don't get how translators would specify other plural type translations that English doesn't need (zero, few, many, etc.) and how our code would know to use them. Do we just define a separate displayMessage for each plural type and just use the same word in English for most of the defaults?

src/data/i18n/default/src/components/AssetsClearFiltersButton/displayMessages.json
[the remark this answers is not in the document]
MichaelRoytman: Ah, I see. Thanks for clarifying.

src/components/AssetsResultsCount/displayMessages.jsx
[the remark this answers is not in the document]
efischer19: I've filed https://openedx.atlassian.net/browse/EDUCATOR-2438 to track the needed i18n work for pluralization, go ahead and merge these messages as they are for now. We'll have to deal with the possibility of seeing `1 possible matches` until that gets resolved.

MichaelRoytman:
I just noticed this. When I'm viewing a course with no assets, the search bar also shows up. Since we're hiding all other filters, I think we should probably hide the search bar, especially since there is nothing to search for.

image

src/components/AssetsPage/displayMessages.jsx
```diff
@@ -0,0 +1,31 @@
...
+    id: 'assetsPageNoResultsNumFiles',
+    defaultMessage: '0 files',
+    description: 'Title for a message displayed when no results are found.',
+  },
+  assetsPageNoResultsMessage: {
+    id: 'assetsPageNoResultsMessage',
+    defaultMessage: 'No files were found.',
+    description: 'Message displayed in table when no results are found.',
+  },
+  assetsPageNoResultsClear: {
```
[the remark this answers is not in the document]
thallada: Oh yes, good catch!

thallada:
@MichaelRoytman good point. The search bar will now only display on a "NORMAL" page (has assets and not filtered down to 0).

I think I addressed all of the comments except for the point about the pluralization which we decided to address in a separate PR later. Let me know if there is anything else, otherwise I think this is good to merge now!

Makefile
```diff
@@ -65,3 +65,6 @@ pull_translations: ## must be exactly this name for edx tooling support, see eco
 
 package-lock.validate:
 	git diff --name-only --exit-code package-lock.json
+
+copy-dist:
+	for f in dist/*; do docker cp $$f edx.devstack.studio:/edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend/dist/; done
```
efischer19: 🙌

PR #149, opened 2018-03-08
i18n message pipeline updates

efischer19:
# [EDUCATOR-2438](https://openedx.atlassian.net/browse/EDUCATOR-2438), part 1

This PR, together with the changes I already merged in https://github.com/edx/studio-frontend/pull/150 and the new transifex output taken from https://github.com/edx/studio-frontend/pull/151 (I can't pull from Transifex natively; the bot has credentials to do so), will greatly simplify the i18n message pipeline.

## Before:
  - messages defined by developers, side-by-side with the relevant components at `src/components/<component>/displayMessages.jsx`
  - `react-intl` default babel extraction then places the messages in `src/data/i18n/default/src/components/<component>/displayMessages.json`, which is tracked in git.
  - `react-intl-gettext` runs `json2po` on that directory, creating `src/data/i18n/default/transifex_input.po`. This file lives in github and is tracked by transifex.
  - translations bot pulls po files into `src/data/i18n/locales/po/<lang>.po`, which are also tracked in github.
  - `react-intl-gettext` runs `po2json` on those directories, creating `src/data/i18n/locales/tmp/<lang>.json` (this is because the`po2json` output is not quite the same as what `react-intl` wants to consume). Note that the `tmp` file is explicitly *not* tracked in github.
  - translations bot then runs `scripts/cleanI18nJson.js` and creates the final, usable translations files at `src/data/i18n/locales/json/<lang>.json`.

## After:
  - messages defined by developers, side-by-side with the relevant components at `src/components/<component>/displayMessages.jsx`
  - `react-intl` default babel extraction then places the messages in `src/data/i18n/default/src/components/<component>/displayMessages.json`. These files are no longer tracked in github, they will be fully generated each time a user (developer, Travis, or translations bot) extracts messages.
  - `reactifex` then runs through those generated files to create a single file, `src/data/i18n/default/transifex_input.json`. This file lives in github and is tracked by transifex.
  - translations bot pulls translations to `src/data/i18n/locales/<lang>.json` directly, and they're already in a format usable by `react-intl`. These translations aren't correct right now, due to Transifex project reconfiguration that I'm also doing to switch from `PO` to `KEYVALUEJSON`.

Developer behavior will not need to change, simply continue to define messages in `displayMessages.jsx` and run `i18n.docker` to update transifex input before merging. Travis will fail your build if you change a string and forget to do this.

Aside from simplifying things, this will allow us to begin uploading ICU formatted messages as discussed in https://github.com/edx/studio-frontend/pull/138#discussion_r171063692. The downside is that we lose developer comments on translation strings, as the `KEYVALUEJSON` format does not support this by default. In part 2 of this work, I'm going to update https://github.com/efischer19/reactifex to also allow for making a series of POST requests to add comments to strings that exist in transifex, as documented here: https://docs.transifex.com/api/resource-strings. Then it will be simple to have the translations bot update comments weekly when it runs the push/pull sequence.


.tx/config
```diff
@@ -1,7 +1,7 @@
 [main]
 host = https://www.transifex.com
 
-[edx-platform.studio-frontend]
-file_filter = src/data/i18n/locales/po/<lang>.po
+[edx-platform.studio-frontend2]
+file_filter = src/data/i18n/locales/proof_of_concept/<lang>.json
 source_lang = en_US
-type = PO
+type = KEYVALUEJSON
```
efischer19: It would be super freaking cool if I could test this out without merging to master, [but apparently that's beyond the capabilities of the transifex bot for some reason](https://github.com/edx/ecommerce-scripts/pull/86)

Will attempt to make more progress tomorrow with judicious use of my admin rights. stay tuned.

efischer19:
The latest commit modifies `make pull_translations` to only pull the set of languages we support; this plus actual translations (the values in transifex needed to be marked "reviewed", that's why all the files are english ATM) will be fully addressed in a follow-up PR. It's included here because I need the bot to run it (has credentials), and the bot cannot run against a branch.

PR #150, opened 2018-03-09
translations job updates

efischer19:
This will greatly simplify the translations pull job


efischer19:
I'm admin merging this; since the translations job can't handle branches and must run against master. The changes should only affect translation work (read- not affect other in flight branches), and I will restore i18n to a working state before we make another release.

I take responsibility if this breaks things.

PR #154, opened 2018-03-09
Transifex bot update translations2018 03 09

efischer19:
This is what I wanted to see:

screen shot 2018-03-09 at 1 41 08 pm


src/data/i18n/locales/currentlySupportedLangs.jsx
```diff
@@ -4,13 +4,10 @@ import esData from 'react-intl/locale-data/es';
 import frData from 'react-intl/locale-data/fr';
 import zhData from 'react-intl/locale-data/zh';
 
-// PUT THESE BACK ONCE THEY EXIST AGAIN
-/*
```
efischer19: This change will make it so that our new message files are included in built output again, restoring master to an "able to be published" state once more.

PR #155, opened 2018-03-12
(EDUCATOR-2164) Refactor Status Alert

MichaelRoytman:


Makefile
```diff
@@ -9,7 +9,7 @@ help: ## display a help message
 shell: ## run a shell on the studio-frontend container
 	docker exec -it dahlia.studio-frontend /bin/bash
 
-attach:
+attach: ## attach local standard input, output, and error streams to studio-frontend container
```
MichaelRoytman: These commands weren't showing up in the output of `make help`, so I thought it might be because they were missing descriptions, but they still don't show up ;(.

efischer19: Are you sure? When I check out your branch, they appear in the help output:

```
ericfischer@Eric-Fischers-MacBook-Pro:~/tmp_code/studio-frontend$ git fetch; gco thallada/refactor-status-alert
remote: Counting objects: 128, done.
remote: Compressing objects: 100% (37/37), done.
remote: Total 128 (delta 83), reused 125 (delta 83), pack-reused 2
Receiving objects: 100% (128/128), 27.87 KiB | 0 bytes/s, done.
Resolving deltas: 100% (83/83), completed with 25 local objects.
From https://github.com/edx/studio-frontend
 + 1f36134...598444e mroytman/hide-image-previews   -> origin/mroytman/hide-image-previews  (forced update)
 + ea0fbd2...7dc5696 thallada/refactor-status-alert -> origin/thallada/refactor-status-alert  (forced update)
Branch thallada/refactor-status-alert set up to track remote branch thallada/refactor-status-alert from origin by rebasing.
Switched to a new branch 'thallada/refactor-status-alert'
ericfischer@Eric-Fischers-MacBook-Pro:~/tmp_code/studio-frontend$ make help
Makefile
Please use `make <target>' where <target> is one of
  asset-page-flag           insert a waffle flag into local docker devstack
  attach                    attach local standard input, output, and error streams to studio-frontend container
  devstack.update           use this if you don't want to fire up the dedicated asset watching containers
  down                      stop and remove studio-frontend container
  from-scratch              start development environment from scratch
  help                      display a help message
  logs                      show logs for studio-frontend container
  npm-install-%             install specified % npm package on the studio-frontend container
  pull_translations         must be exactly this name for edx tooling support, see ecommerce-scripts/transifex/pull.py
  restart-detached          bring container down and back up in detached mode
  restart                   bring container down and back up
  shell                     run a shell on the studio-frontend container
  up-detached               bring up studio-frontend container in detached mode
  up                        bring up studio-frontend container
```

As far as I can tell, the perl regex in `help`'s definition is explicitly checking for the `##` divider to use as a description in the output

src/components/AssetsPage/index.jsx
```diff
@@ -29,10 +30,18 @@ export default class AssetsPage extends React.Component {
...
+    this.statusAlertRef = null;
+    this.deleteButtonRefs = {};
+
+    this.getNextFocusElementOnDelete = this.getNextFocusElementOnDelete.bind(this);
+    this.onDeleteStatusAlertClose = this.onDeleteStatusAlertClose.bind(this);
   }
 
   componentDidMount() {
-    this.props.getAssets({}, this.props.courseDetails);
+    if (this.props.assetsList.length === 0) {
```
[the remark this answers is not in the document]
efischer19: Smart!

src/components/AssetsPage/index.jsx
```diff
@@ -213,4 +254,9 @@ AssetsPage.propTypes = {
   searchSettings: PropTypes.shape({
     enabled: PropTypes.bool,
   }).isRequired,
+  clearAssetDeletion: PropTypes.func.isRequired,
+};
+
+AssetsPage.defaultProps = {
+  deletedAssetIndex: null,
```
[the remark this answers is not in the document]
efischer19: This smells funny. I get why you had to do it, but the fact that it was needed is still.. ugh.

Is there a way to force the Redux default to a number?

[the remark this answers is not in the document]
MichaelRoytman: The actual default in Redux is null already, so this is just mirroring that. Is that okay? Or would you prefer that the null value for this variable be a number everywhere? We could do -1 as our null value and just check that deletedAssetIndex is equal to/not equal to -1 where it matters. What do you think?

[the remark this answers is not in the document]
efischer19: I'm not a big fan of special-cases for null in general, but if Redux is already using `null` as a default I think that's the way to go over adding our own custom -1 logic. At least then it's clearly a "this was never set" value instead of the result of an overzealous decrement, for example.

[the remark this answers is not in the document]
MichaelRoytman: Sorry, I think I wasn't super clear. I'm the one that set it to `null` [here](https://github.com/edx/studio-frontend/pull/155/files#diff-1dcfcb67a72beb3a12a086472ae272aeR15). Does that change your opinion?  If I were to omit the initialization to `null`, I believe that the default value would be `undefined` instead. There is a Number object in Javascript we could leverage, perhaps (i.e. `null` might make more sense as the default for an object than for a primitive).

src/components/AssetsPage/index.jsx
```diff
@@ -41,6 +50,23 @@ export default class AssetsPage extends React.Component {
...
+    // do not attempt to focus if there are no assets delete buttons to focus on
+    // TO-DO: determine where the focus should go when the last asset is deleted
+    if (this.props.assetsList.length > 0) {
+      const focusElement = this.getNextFocusElementOnDelete();
+      focusElement.focus();
+    }
+  }
+
+  getNextFocusElementOnDelete() {
+    const deletedIndex = this.props.deletedAssetIndex;
```
jaebradley: Nit: you could consider eliminating this variable (I don't know if you run into line-length problems)

```javascript
const assetToDelete = this.props.assetsList[this.props.deletedAssetIndex];
```

If you run into line-length issues, maybe

```javascript
const { deletedAssetIndex, assetsList } = this.props;

return this.deleteButtonRefs[assetsList[deletedAssetIndex].id];
```

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+    this.statusAlertRef = {};
+
+    this.closeStatusAlert = this.closeStatusAlert.bind(this);
+    this.updateStatusAlertFields = this.updateStatusAlertFields.bind(this);
+    this.updateUploadSuccessCount = this.updateUploadSuccessCount.bind(this);
+
+    this.closeDeleteStatus = this.closeDeleteStatus.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
```
jaebradley: I think you can destructure in `componentWillReceiveProps` directly?

```javascript
componentWillReceiveProps({ assetsStatus, deletedAsset }) {
  this.updateStatusAlertFields(assetsStatus, deletedAsset);
}
```

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+      case assetActions.lock.TOGGLING_LOCK_ASSET_FAILURE:
+        alertDialog = (
+          <WrappedMessage
+            message={messages.assetsStatusAlertFailedLock}
+            values={{ assetName: assetsStatus.asset.name }}
+          />
+        );
+        alertType = 'danger';
+        break;
+      case assetActions.clear.CLEAR_FILTERS_FAILURE:
```
jaebradley: Can these next four `case`s be combined?

```javascript
case assetActions.clear.CLEAR_FILTERS_FAILURE:
case assetActions.filter.FILTER_UPDATE_FAILURE:
```

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+    let alertDialog;
+    let alertType;
+
+    const genericUpdateError = (
+      <WrappedMessage
+        message={messages.assetsStatusAlertGenericUpdateError}
+      />
+    );
+
+    switch (assetsStatus.type) {
```
jaebradley: It feels like this `switch` could be moved out into it's own function since what it's really doing is generating two pieces of information that will eventually get used in the `statusAlertFields` update?

Maybe something like

```javascript
getStatusAlertFields(assetStatus) {
  // add switch here
  ...
  return { alertDialog, alertType }
}

updateStatusAlertFields(assetsStatus, deletedAsset) {
  this.setState({
    statusAlertOpen: true,
    statusAlertFields: this.getStatusAlertFields(assetStatus),
  });
}
```

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+  closeDeleteStatus = () => {
+    this.closeStatusAlert();
+    this.props.onDeleteStatusAlertClose();
+  }
+
+  closeStatusAlert() {
+    this.props.clearAssetsStatus();
+
+    // clear out all status related state
+    this.setState({
```
jaebradley: This looks the same as the default `state`? You could consider moving this into some constant instead of repeating?

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+      statusAlertFields: {
+        alertDialog: '',
+        alertType: 'info',
+      },
+      uploadSuccessCount: 1,
+    });
+  }
+
+  updateUploadSuccessCount() {
+    const uploadSuccessCount = this.state.uploadSuccessCount + 1;
```
jaebradley: Nit: you could consider forgoing the variable assignment like

```javascript
this.setState({ uploadSuccessCount: this.state.uploadSuccessCount + 1 });
```

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+    this.setState({
+      uploadSuccessCount,
+    });
+  }
+
+  render() {
+    const { assetsStatus } = this.props;
```
jaebradley: You could consider adding `statusAlertRef` to the destructuring here.

jaebradley: You could also consider adding `this.state` destructuring as well

src/components/AssetsStatusAlert/index.jsx
```diff
@@ -0,0 +1,228 @@
...
+
+    let onClose = this.closeStatusAlert;
+    if (Object.prototype.hasOwnProperty.call(assetActions.delete, assetsStatus.type)) {
+      onClose = this.closeDeleteStatus;
+    }
+
+    const statusAlert = (
```
jaebradley: You could consider `return`ing directly here instead of assigning to a variable and then `return`ing that variable.

Makefile
MichaelRoytman: That's odd. This is what I see.

```
mroytman:~/workspaces/studio-frontend $ make help
Makefile
Please use `make <target>' where <target> is one of
  asset-page-flag           insert a waffle flag into local docker devstack
  devstack.update           use this if you don't want to fire up the dedicated asset watching containers
  down                      stop and remove studio-frontend container
  from-scratch              start development environment from scratch
  help                      display a help message
  logs                      show logs for studio-frontend container
  npm-install-%             install specified % npm package on the studio-frontend container
  pull_translations         must be exactly this name for edx tooling support, see ecommerce-scripts/transifex/pull.py
  shell                     run a shell on the studio-frontend container
  up-detached               bring up studio-frontend container in detached mode
  up                        bring up studio-frontend container
```

src/components/AssetsStatusAlert/index.jsx
MichaelRoytman: Yes. Thank you. I tried to figure out how to combine cases but wasn't able to find the correct way to do it online.

src/components/AssetsPage/index.jsx
MichaelRoytman: I think I prefer keeping focusAsset. I think it makes it the code clearer, instead of indexing into two separate lists, but I like the prop destructuring.

src/components/AssetsPage/index.jsx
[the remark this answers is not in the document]
efischer19: Short answer: go ahead and merge this, `null` is alright here. The funny smell I referenced above is likely not something we can easily address at this time.

Longer answer: The concept of a null value in general is something I don't particularly like, for reasons that are fairly well summarized [here](https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/). Further, the fact that javascript makes it worse by having both `undefined` and `null` exist is pants-on-head dumb, similar to the famous `===` in terms of "things that should never have had reason to exist". IMO, the "cleanest" possible way to code this would be to avoid null/sentinel/hasn't-been-set-yet values at all. As a thought experiment say we split `deletedAssetIndex` into two variables `somethingWasDeletedJustNow` (boolean, never null) and `deletedAssetIndex` (number, never null). Wouldn't the propTypes be simpler to handle throughout now that we have a boolean and a number? Optionals/Maybes attempt to solve this (see https://www.npmjs.com/package/maybe as an example), but are well beyond the scope of this change.

PR #159, opened 2018-03-20
Bump to 1.6.1, upgrade Paragon

thallada:
Upgrade Paragon to 2.4.1, which fixes studio-frontend in IE11

Bumping the minor version to cover @MichaelRoytman new hide image preview option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "1.5.0",
+  "version": "1.6.1",
   "description": "The frontend for the Open edX platform",
   "repository": "edx/studio-frontend",
   "scripts": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@edx/studio-frontend",
-  "version": "1.5.0",
+  "version": "1.6.1",
```
MichaelRoytman: I was under the impression that when you change the minor version, the patch version gets reset to 0? Is that incorrect?

https://semver.org/#spec-item-7

thallada: My reasoning was: 

1. bump to v1.6.0 for the preview image toggle (we never released that)
2. bump to v1.6.1 for the IE11 fix and the Mac Firefox searchfield fix

I can make it 1.6.0 if you think that makes more sense though.

MichaelRoytman: Ah, I see. That makes sense. Thanks. I'll approve it now.

PR #160, opened 2018-03-22
Add travisbuddy to .travis.yml

thallada:
Travis buddy is a bot that will comment on PRs letting you know what to fix when tests fail.

I'm putting this PR up as a test. Let me know if you think it is annoying/not useful.


thallada:
Implemented now :)

PR #162, opened 2018-03-23
Upgrade to webpack 4

thallada:
Upgrade to [Webpack 4](https://medium.com/webpack/webpack-4-released-today-6cdb994702d4).


config/webpack.prod.config.js
```diff
@@ -87,15 +86,19 @@ module.exports = Merge.smart(commonConfig, {
       },
     ],
   },
+  optimization: {
+    runtimeChunk: 'single',
+    splitChunks: {
+      cacheGroups: {
+        commons: {
+          name: 'common',
```
thallada: Instead of putting all node_module packages into a `vendor.min.js`, this will put any chunk that is common in more than one entry into a `common.min.js`. This will require a change in the platform since we will need to load `common.min.js` instead of `vendor.min.js`.

config/webpack.prod.config.js
```diff
@@ -106,6 +109,5 @@ module.exports = Merge.smart(commonConfig, {
         NODE_ENV: JSON.stringify(process.env.NODE_ENV),
       },
     }),
-    new UglifyJsPlugin({ sourceMap: true }),
```
thallada: Done automatically by webpack in the production mode.

config/webpack.prod.config.js
```diff
@@ -87,15 +86,19 @@ module.exports = Merge.smart(commonConfig, {
...
+  },
   plugins: [
-    new webpack.optimize.CommonsChunkPlugin({
-      name: 'vendor',
-      minChunks: module => module.context && module.context.includes('node_modules'),
-    }),
-    new webpack.optimize.CommonsChunkPlugin({
-      name: 'manifest',
-      minChunks: Infinity,
-    }),
```
thallada: CommonChunkPlugin was replaced by the optimization.splitChunks option.

config/webpack.common.config.js
```diff
@@ -9,7 +9,6 @@ module.exports = {
     i18nMessages: path.resolve(__dirname, '../src/data/i18n/locales/currentlySupportedLangs.jsx'),
   },
   output: {
-    filename: '[name].js',
```
thallada: This the default already.

package.json
```diff
@@ -42,43 +42,43 @@
...
     "babel-preset-env": "^1.6.1",
     "babel-preset-react": "^6.24.1",
     "codecov": "^2.3.0",
-    "css-loader": "^0.28.5",
+    "css-loader": "^0.28.11",
     "enzyme": "^3.1.1",
     "enzyme-adapter-react-16": "^1.0.4",
     "eslint-config-edx": "4.0.0",
-    "extract-text-webpack-plugin": "^3.0.0",
+    "extract-text-webpack-plugin": "next",
```
thallada: This package hasn't quite released support for webpack 4 yet, but I'm tracking it and will update this to a version number when it does.

efischer19: does this mean we're blocked on merging this?

answering my own question: nope, because `next` maps to `4.0.0-beta.0` according to https://www.npmjs.com/package/extract-text-webpack-plugin?activeTab=versions

PR #163, opened 2018-03-27
Export language files

efischer19:
# [EDUCATOR-2584](https://openedx.atlassian.net/browse/EDUCATOR-2584)

## Fix for non-exported message files

Introduced via c8586bc, I should
have updated the path in this file at that time as well.

### DIFF
--- a/config/webpack.prod.config.js
+++ b/config/webpack.prod.config.js
@@ -26,9 +26,10 @@ module.exports = Merge.smart(commonConfig, {
         loader: 'babel-loader',
       },
       {
+        type: 'javascript/auto',
         test: /\.json$/,
         include: [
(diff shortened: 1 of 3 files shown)


efischer19:
I just rebased to grab @thallada's changes before releasing `1.6.2`

config/webpack.prod.config.js
```diff
@@ -26,9 +26,10 @@ module.exports = Merge.smart(commonConfig, {
         loader: 'babel-loader',
       },
       {
+        type: 'javascript/auto',
```
efischer19: This was the default for everything in webpack 3, we now need to add it explicitly to prevent webpack from "helpfully" doing the wrong thing with json files.

PR #165, opened 2018-03-29
Finalize i18n pipeline upgrades

efischer19:
# [EDUCATOR-2438](https://openedx.atlassian.net/browse/EDUCATOR-2438)

With these changes, we have fully migrated from using the `PO` filetype on Transifex to the `KEYVALUEJSON` filetype, which has proper support for pluralization. The addition of reactifex 1.0.0 also allows us to keep translator comments, which are not included in `KEYVALUEJSON` by default.


src/data/i18n/default/transifex_input.json
```diff
@@ -50,8 +50,8 @@
   "assetsPageNoResultsMessage": "No files were found.",
   "assetsPageNoAssetsNumFiles": "0 files in your course",
   "assetsPageNoAssetsMessage": "Enhance your course content by uploading files such as images and documents.",
-  "assetsResultsCountFiltered": "Showing {start}-{end} out of {total} possible matches.",
-  "assetsResultsCountTotal": "Showing {start}-{end} out of {total} total files.",
+  "assetsResultsCountFiltered": "Showing {start}-{end} out of {total, plural, one {1 possible match} other {{total} possible matches}}.",
+  "assetsResultsCountTotal": "Showing {start}-{end} out of {total, plural, one {1 total file} other {{total} total files}}.",
```
efischer19: I think this format matches [Transifex specification for plurals](https://docs.transifex.com/formats/json#plurals-support)

src/data/i18n/default/src/components/AssetsImagePreviewFilter/displayMessages.json
```diff
@@ -1,7 +0,0 @@
-[
```
efischer19: These two files were added before my `.gitignore` change affected them; we no longer store them in github (they're calculated on-the-fly) as of https://github.com/edx/studio-frontend/pull/149

Makefile
```diff
@@ -60,6 +60,15 @@ i18n.preprocess: ## gather all display strings into a single file
 i18n.pre_validate: | i18n.extract i18n.preprocess
 	git diff --exit-code ./src/data/i18n/default/transifex_input.json
 
+tx_url1 = https://www.transifex.com/api/2/project/edx-platform/resource/studio-frontend/translation/en/strings/
+tx_url2 = https://www.transifex.com/api/2/project/edx-platform/resource/studio-frontend/source/
+push_translations: | i18n.extract
+	# Transifex is set up to watch transifex_input.json, but that filetype lacks comments
+	./node_modules/reactifex/bash_scripts/get_hashed_strings.sh $(tx_url1)
+	$$(npm bin)/reactifex ./src/data/i18n/default/src/components/ --comments
+	./node_modules/reactifex/bash_scripts/put_comments.sh $(tx_url2)
```
efischer19: See https://github.com/efischer19/reactifex/blob/master/README.md for details

thallada: Cool package, but I can't help but think that JS dynamically creating a bash script seems code smelly. You mentioned that only curl was working for you... what was wrong with doing the request in Node instead?

efischer19: Agreed, it wasn't my first choice.

The problem came in when trying to get basic auth to succeed while handling 301 redirects ([Transifex docs)(https://docs.transifex.com/api/introduction#authentication)). I couldn't get [the default HTTPS Node library](https://nodejs.org/api/https.html) to do that correctly, so I defaulted back to the documented-by-Transifex approach of using curl.

It'd arguably be better to add a dependency on `[request](https://www.npmjs.com/package/request)` and use that; it just felt a little heavyweight compared to the dead-simple curls outlined in the transifex docs.

I just added some tests with an example of `put_comments.sh`, if seeing the results makes things any clearer. All the js does is match up message ids with Transifex-defined `string_hash` values to construct a URL. https://github.com/efischer19/reactifex/blob/master/bash_scripts/put_comments.sh

efischer19: as of the 1.0.2 release, the above linked script is now static while the *data it consumes* is dynamic, which seems way better.

efischer19:
Updated to get these screenshots locally, we need to make sure that a raw number is passed into these plural strings (as opposed to the results of a `renderCount()` call)

screen shot 2018-04-03 at 11 51 09 am

screen shot 2018-04-03 at 11 51 16 am

src/components/AssetsResultsCount/index.jsx
```diff
@@ -27,7 +27,8 @@ const AssetsResultsCount = ({ paginationMetadata, filtersMetadata, searchMetadat
         values={{
           start: renderCount(paginationMetadata.start + 1),
           end: renderCount(paginationMetadata.end),
-          total: renderCount(paginationMetadata.totalCount),
+          formatted_total: renderCount(paginationMetadata.totalCount),
+          total: paginationMetadata.totalCount,
```
efischer19: This is the key - both are variables passed through to react-intl, but one is a number while the other is a span containing a number

PR #166, opened 2018-04-03
add extract_translations make target

efischer19:
https://github.com/edx/ecommerce-scripts/blob/27b3f675db2e2f0b1fd2ac2fe148589e94909ff7/transifex/push.py#L34

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ asset-page-flag: ## insert a waffle flag into local docker devstack
 i18n.docker: ## what devs should do from their host machines
 	docker exec -t dahlia.studio-frontend bash -c 'make i18n.extract && make i18n.preprocess'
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -51,6 +51,10 @@ asset-page-flag: ## insert a waffle flag into local docker devstack
 i18n.docker: ## what devs should do from their host machines
 	docker exec -t dahlia.studio-frontend bash -c 'make i18n.extract && make i18n.preprocess'
 
+extract_translations: | i18n.extract ## tooling expects a target with this name
+	# every other repo is python, so we need to use this hook to npm install on jenkins
+	npm install
```
thallada: In the future, as we have more JS-only IDAs, we might want to change the jenkins job so it knows to `npm install` for JS repos, but for now I'm alright with this.

efischer19: agreed

PR #167, opened 2018-04-04
(EDUCATOR-2594) SFE.scss fixup

MichaelRoytman:


src/components/AssetsPage/index.jsx
```diff
@@ -110,34 +110,34 @@ export default class AssetsPage extends React.Component {
...
-      <div className={edxBootstrap['col-10']}>
-        <div className={edxBootstrap.row}>
-          <div className={`${edxBootstrap['col-md-8']}`}>
+      <div className="col-10">
+        <div className="row">
+          <div className="col-md-8">
             <WrappedAssetsResultsCount />
           </div>
-          <div className={`${edxBootstrap['col-md-4']} ${edxBootstrap['text-right']}`}>
+          <div className={classNames('col-md-4', 'text-right')}>
```
MichaelRoytman: Are we still using classnames? Or should I just do the {[ .. ]} syntax?

src/components/AccessibilityBody/AccessibilityBody.scss
```diff
@@ -1,21 +1,23 @@
 @import 'edx-bootstrap';
 
-.list {
-  $bottom-spacer: $spacer * 0.5;
-  margin-left: $spacer * 2;
-  margin-bottom: $bottom-spacer;
-
-  li {
+.SFE-wrapper {
```
MichaelRoytman: Since all custom styles have been wrapped in the SFE-wrapper scss class, do we have to change how we use them when the file is imported?

For example, would it be  

`<ol className={styles['numerical-list']}>` or `<ol className={styles['SFE-wrapper']['numerical-list']}>`?

I thought it would be the second, but I got some failing tests with classNames being undefined, so I think I was mistaken. Just want to double-check and understand why/how they're imported.

thallada: Oh darn. I don't think CSS modules is compatible with Sass nesting like this. I think we have a few options:

1) Just import styles like `import './XComponent.scss';` and then reference the classes in it by string name without css modules: `className="list"`. This runs the risk that multiple styles with the same class name could conflict between our different components, but IMO that's a pretty small risk.
2) Prefix every selector with `.SFE-wrapper ` instead. This seems more annoying than wrapping everything in one `.SFE-wrapper` block. I think CSS modules would work better with that, but I didn't test it.
3) I tried changing our webpack config to automatically prepend `#root.SFE .SFE-wrapper` to everything except for stuff in `src/SFE.scss` and `font-awesome.min.css` (which it will prepend `#root.SFE` to instead), but I can't seem to get it to work in the dev config and it's pretty ugly/complicated because it requires me to duplicate the same rule twice but with a tiny tweak to one postcss plugin and different exclude/include options.
4) Use an iframe(??). Obviously a bigger change, but the CSS conflicts are the reason for all of this complexity and I'm wondering if we didn't investigate that option fully enough.

src/components/AssetsPage/index.jsx
thallada: I think this is basically equivalent to `<div className="col-md-4 text-right">`. Using classNames is useful if we want to change classes based off of variables, so in this case I think a plain string is simpler.

MichaelRoytman:
@thallada Do you mind taking another look at this?

PR #174, opened 2018-04-06
(EDUCATOR-2486) field validation and edit image modal status alert

MichaelRoytman:


src/components/EditImageModal/index.jsx
```diff
@@ -119,22 +126,39 @@ export default class EditImageModal extends React.Component {
...
-        detail: {
-          height: this.getDimensionStateOrNatural('height'),
-          width: this.getDimensionStateOrNatural('width'),
-          src: this.state.imageSource,
-          alt: this.state.isImageDecorative ? '' : this.state.imageDescription,
-          style: this.state.imageStyle,
-        },
-      },
-    ));
+    const isValidImageSourceURL = { isValid: true };
```
MichaelRoytman: This is a placeholder for image source URL validation.

src/components/EditImageModal/index.jsx
```diff
@@ -19,23 +22,31 @@ export default class EditImageModal extends React.Component {
...
     this.onImageDescriptionBlur = this.onImageDescriptionBlur.bind(this);
     this.onInsertImageButtonClick = this.onInsertImageButtonClick.bind(this);
     this.onImageIsDecorativeClick = this.onImageIsDecorativeClick.bind(this);
     this.onImageLoad = this.onImageLoad.bind(this);
     this.onImageSourceBlur = this.onImageSourceBlur.bind(this);
+    this.onStatusAlertClose = this.onStatusAlertClose.bind(this);
+    this.validateImageDescription = this.validateImageDescription.bind(this);
 
-    this.formRef = {};
+    this.formRef = null;
```
MichaelRoytman: I think the initial value should be null, per [this](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) React documentation.

src/SFE.scss
```diff
@@ -45,6 +45,11 @@
   border-bottom: 1px solid theme-color(light);
 }
 
+.bullet-list {
```
MichaelRoytman: Pulling some of these blocks out of the accessibility policy page into here, so they can be used both in the accessibility policy page and in the edit image modal.

src/_var-reset.scss
```diff
@@ -77,3 +77,4 @@ $breadcrumb-padding-x: $breadcrumb-padding-x / $base-rem-size;
 $breadcrumb-item-padding: $breadcrumb-item-padding / $base-rem-size;
 $kbd-padding-x: $kbd-padding-x / $base-rem-size;
 $kbd-padding-y: $kbd-padding-y / $base-rem-size;
+$close-font-size: $font-size-base * 1.5;
```
MichaelRoytman: @fysheets contributed this to make the x button in the modal/status alert look better.

src/components/EditImageModal/EditImageModal.scss
```diff
@@ -1,74 +1,87 @@
...
-    @extend .mb-4;
-  }
+fieldset {
+  @extend .mb-4;
+}
 
-  img {
-    max-width: 100%;
-  }
+.image-preview-container {
```
MichaelRoytman: I just made some of the selectors more descriptive.

src/components/EditImageModal/EditImageModal.scss
```diff
@@ -1,74 +1,87 @@
...
+  position: relative;
+  width: 100%;
+}
 
-  .image-preview {
-    @extend .p-3;
-    max-width: 100%;
-    text-align: center;
-    background-color: theme-color('lightest');
+.image-preview-image {
```
MichaelRoytman: This centers the image inside of its parent (the placeholder div).

src/components/EditImageModal/EditImageModal.scss
```diff
@@ -1,74 +1,87 @@
 @import 'edx-bootstrap';
 
-.SFE-wrapper {
```
MichaelRoytman: I'm undoing this nesting here, since we figured out how to avoid it, hence why a lot of this file has been modified because of indentation.

src/components/EditImageModal/index.jsx
```diff
@@ -1,44 +1,62 @@
...
 import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
 import rewriteStaticLinks from '../../utils/rewriteStaticLinks';
 
 const LOADING_SPINNER_DELAY = 1000; // in milliseconds
 
+const imageDescriptionHelpLink = 'http://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/accessibility/best_practices_course_content_dev.html#use-best-practices-for-describing-images';
+const imageDescriptionID = 'imageDescription';
+const imageSourceID = 'imageSource';
+
+const initialEditImageModalState = {
```
MichaelRoytman: This allows an easier reset of state when modal is opened via the "handleModalOpen" function.

src/components/EditImageModal/index.jsx
```diff
@@ -342,29 +417,78 @@ export default class EditImageModal extends React.Component {
   getImage = () => (
     <img
       alt=""
+      className={classNames(styles['image-preview-image'], { invisible: !this.state.isImageLoaded })}
```
MichaelRoytman: If the image is not loaded, hide the img element.

src/components/EditImageModal/index.jsx
```diff
@@ -342,29 +417,78 @@ export default class EditImageModal extends React.Component {
...
+  /*
+   image is conditionally displayed so that onImageError is not called on initial
+    modal open, when image source is empty string
+  */
   getImagePreviewPlaceholder = () => (
-    <div className="image-preview"><WrappedMessage message={messages.editImageModalImagePreviewText} /></div>
+    <div className={styles['image-preview-placeholder']}>
+      <WrappedMessage message={messages.editImageModalImagePreviewText} >
+        {displayText =>
+          (<span className={classNames({ invisible: this.state.isImageLoaded })}>
```
MichaelRoytman: We hide the "Image Preview" text inside of the placeholder div if an image is loaded over top of it, to avoid it peeking out from underneath.

src/components/EditImageModal/index.jsx
```diff
@@ -342,29 +417,78 @@ export default class EditImageModal extends React.Component {
...
+            {displayText}
+          </span>)}
+      </WrappedMessage>
+      { this.state.imageSource && this.getImage() }
```
MichaelRoytman: See code comment above to explain why I'm doing this.

src/components/EditImageModal/index.jsx
```diff
@@ -404,10 +530,46 @@ export default class EditImageModal extends React.Component {
       imageSource: event.detail.src || '',
       imageStyle: event.detail.style || '',
       isImageDecorative: event.detail.alt === '',
+      // if existing img had a source, assume it could be loaded and show the image preview
```
MichaelRoytman: I'm not sure whether this will cause some funkiness with an edge case I couldn't think of, but it's the best way I could ensure that an image preview is shown again when editing an existing image.

MichaelRoytman:
TODO:
- Once the paragon Fieldset component is merged, I'll have to modify this to use that component and ensure it can be validated.

src/components/EditImageModal/EditImageModal.scss
fysheets: If you add `?w=1` to the end of the URL it won't show any of the whitespace changes :)

src/components/EditImageModal/index.jsx
```diff
@@ -342,29 +416,78 @@ export default class EditImageModal extends React.Component {
...
+    </div>
+  );
+
+  getStatusAlertDialog = () => (
+    <div>
+      <WrappedMessage
+        message={messages.editImageModalFormErrorMissingFields}
+        tagName="div"
+      />
+      <br />
```
fysheets: We should use styling here instead of the `<br />`. 

[The br element may only be used for line breaks that are actually part of the content.](https://www.w3.org/TR/2014/REC-html5-20141028/text-level-semantics.html#the-br-element) We want this to be a visual indicator of space between the message and the list, so it would be better practice to add a class to adjust the margin/padding instead of using this tag.

src/components/EditImageModal/index.jsx
```diff
@@ -342,29 +416,78 @@ export default class EditImageModal extends React.Component {
...
+      <div>
+        <ul className="bullet-list">
+          {(Object.keys(this.state.validationMessages).reduce((accumulator, current) => {
+            const value = this.state.validationMessages[current];
```
fysheets: Could this be `currentValidationMessages` or something more descriptive?

src/components/EditImageModal/index.jsx
```diff
@@ -404,10 +529,46 @@ export default class EditImageModal extends React.Component {
       imageSource: event.detail.src || '',
       imageStyle: event.detail.style || '',
       isImageDecorative: event.detail.alt === '',
+      // if existing img had a source, assume it could be loaded and show the image preview
+      isImageLoaded: !!event.detail.src,
```
fysheets: Doesn't the `!!` cancel itself out? Could we just remove those?

fysheets: Nevermind :) Picked @arizzitano's brain

src/components/EditImageModal/index.jsx
MichaelRoytman: That sounds good. Thank you for correcting that. I copied it from the [accessibility page|https://github.com/edx/studio-frontend/blob/25bd9cefb27514511be7c23153696baa767f3b9e/src/components/AccessibilityPolicyForm/index.jsx#L94], so we should be sure to make that change there too.

src/components/EditImageModal/index.jsx
MichaelRoytman: Yeah, it's a weird thing :P.

src/components/EditImageModal/index.jsx
```diff
@@ -1,44 +1,61 @@
...
 
 import messages from './displayMessages';
-import './EditImageModal.scss';
+import styles from './EditImageModal.scss';
 import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
 import rewriteStaticLinks from '../../utils/rewriteStaticLinks';
 
 const LOADING_SPINNER_DELAY = 1000; // in milliseconds
 
+const imageDescriptionHelpLink = 'http://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/accessibility/best_practices_course_content_dev.html#use-best-practices-for-describing-images';
```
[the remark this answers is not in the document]
MichaelRoytman: Ah, okay. That sounds good. The one thing I'm not sure of is that we're not using Redux for this component, which is how we access the studioContext elsewhere. The accessibility page doesn't subscribe to any Redux store either, and it gets necessary URLs via the accessibilityIndex.jsx file, which passes in the URL(s) as a [prop(s)](https://github.com/edx/studio-frontend/blob/master/src/accessibilityIndex.jsx#L36). Does that seem like an analogous situation to this one? Or should I figure out a way to get this documentation URL into this component through Redux?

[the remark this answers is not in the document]
dsjen: Hmm, I think the studio context is a global at this point, so that might be a way to access it--don't love that solution. Is the edit modal instantiated in Studio's mako template? Perhaps the prop pattern could be used here and Studio would be updated to pass along the prop. Does that work?

MichaelRoytman:
[This edx-platform PR](https://github.com/edx/edx-platform/pull/18006#issuecomment-383168845) will make the image accessibility documentation available through the studioContext global.

PR #177, opened 2018-04-09
Add URL input validation, update Bootstrap

thallada:
Also uses bootstrap input groups to add a loading spinner to the URL input when the image is loading.

screenshot-20180409122119-648x162

screenshot-20180409122132-650x187


src/_var-reset.scss
```diff
@@ -78,5 +75,5 @@ $thumbnail-padding: $thumbnail-padding / $base-rem-size;
 $breadcrumb-padding-y: $breadcrumb-padding-y / $base-rem-size;
 $breadcrumb-padding-x: $breadcrumb-padding-x / $base-rem-size;
 $breadcrumb-item-padding: $breadcrumb-item-padding / $base-rem-size;
-$code-padding-x: $code-padding-x / $base-rem-size;
-$code-padding-y: $code-padding-y / $base-rem-size;
```
thallada: These bootstrap variables were removed/renamed in the latest Bootstrap v4.0.0

src/components/EditImageModal/EditImageModal.test.jsx
```diff
@@ -30,7 +30,7 @@ const sampleImgData = {
 };
 sampleImgData.aspectRatio = sampleImgData.naturalWidth / sampleImgData.naturalHeight;
 
-const updateConsants = () => {
```
thallada: I think this name was typo-ed so I did a find-and-replace to fix it in this file.

src/editImageModalIndex.jsx
```diff
@@ -10,6 +10,7 @@ if ((typeof window !== 'undefined' && !window._babelPolyfill) ||
 import React from 'react';
 import ReactDOM from 'react-dom';
 import { IntlProvider } from 'react-intl';
+import 'custom-event-polyfill'; // needed for IE11
```
thallada: This is kind of unrelated to this PR, because I noticed that the new modal is broken in IE11 on master right now. This import fixes it though.

src/components/EditImageModal/EditImageModal.test.jsx
MichaelRoytman: My bad. Thank you.

src/components/EditImageModal/EditImageModal.test.jsx
```diff
@@ -576,7 +623,7 @@ describe('EditImageModal', () => {
       it('width responds to height change', () => {
         imageDimensionsHeightInput.find('input').simulate('change', { target: { value: '400' } });
 
-        updateConsants();
+        updateConstants();
 
         // issue with props in asInput expecting a string even if type is number
```
MichaelRoytman: I just noticed this comment I think I left. I think this isn't an issue anymore, right? Should we remove this comment?

src/components/EditImageModal/index.jsx
```diff
@@ -68,19 +69,22 @@ export default class EditImageModal extends React.Component {
...
       imageLoading: false,
+      isImageValid: false,
     });
   }
 
   onImageSourceBlur = (imageSource) => {
     this.setState({
       imageSource,
-      imageLoading: imageSource.length > 0,
+      imageLoading: imageSource.length > 0 && (this.state.imageSource !== imageSource),
```
MichaelRoytman: I'm wondering whether we should have more states for this image? I'm finding that I'm going to need some notion of "imageIsEmpty" and "imageIsLoaded" so that I know when to conditionally hide the placeholder div once the image is loaded. Would me adding that in my pull request conflict with anything you've done here?

src/editImageModalIndex.jsx
MichaelRoytman: 👍

MichaelRoytman:
Oh, some other things I wanted to mention that are definitely not required but some things I thought about:

1) Is there a way to keep the loading spinner from flickering like that? I think the loader makes sense for slow network requests, but, otherwise, it's kind of distracting when it flashes on the screen. I don't know what the best course of action would be. Maybe we create an artificial delay of 1 second so that the spinner shows for 1 second and then disappears? 1 second seems not too detrimental. Or maybe we show it if the delay has taken longer than 1 second? Obviously, we can't anticipate the request duration, so this would at least give us some hurdle to jump over for slow requests.

2) Is it worth catching the error that gets logged to console when an invalid image URL is entered so that it's not shown to the user, since we're already addressing it in our status alert/validation?

src/components/EditImageModal/displayMessages.jsx
```diff
@@ -81,6 +81,21 @@ const messages = defineMessages({
     defaultMessage: 'or',
     description: 'Describes two choices,',
   },
+  editImageModalImageNotFoundError: {
+    id: 'editImageModalImageNotFoundError',
+    defaultMessage: 'Image could not be loaded from this URL',
+    description: 'Error message displayed below URL field input when user-entered URL did not load an image',
+  },
+  editImageModalImageNotFoundDangerIconDescription: {
```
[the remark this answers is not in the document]
thallada: That makes sense if we are going to reuse this same message in multiple places. I'll make that change, thanks.

src/SFE.scss
```diff
@@ -55,3 +55,9 @@ button {
 input[type="search"] {
   appearance: none;
 }
+
+// Paragon-specific styling copied here from:
+// https://github.com/edx/paragon/blob/efd86c59fe6266b569939a2d5f526fce5f2ecafc/src/asInput/asInput.scss#L5-L7
+.fa-icon-spacing {
```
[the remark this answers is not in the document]
thallada: Like I explained in-person, I think we should start using the CSS generated by Paragon's themeable build so we don't have to keep copy-and-pasting CSS from Paragon. It includes Bootstrap and Font Awesome, so we will have to do some refactoring to use those from Paragon instead of importing it separately in studio-frontend. In the interest of getting this merged soon, I created a separate ticket to do that work: https://openedx.atlassian.net/browse/EDUCATOR-2704

src/components/EditImageModal/index.jsx
thallada: I think as long as `imageLoading` is still set when the image is loading then everything here should still work.

src/components/EditImageModal/EditImageModal.test.jsx
thallada: I changed the values in the tests with this comment to numbers and the tests pass with no errors/warnings, so I think it is resolved. I deleted these comments.

thallada:
@MichaelRoytman I think I addressed most of your comments.

1. That's a great idea. I added a 1 second timeout that will display the spinner if the image is still loading.
2. Unfortunately, I don't think that is possible since the console error is getting emitted from the `` tag in the browser. [StackOverflow says it is unavoidable unless a server-side request makes the check](https://stackoverflow.com/a/7041565/6620612).

MichaelRoytman:
Thanks for addressing the comments! Too bad about the console error, but it makes sense :).

PR #181, opened 2018-04-12
Wrap long filenames in the AssetsTable

thallada:
Thanks @MichaelRoytman for spotting this bug. I also removed the `.wrap-text` span around the table because I don't think it was doing anything to wrap text.


fysheets:
@thallada validated locally 👍

PR #183, opened 2018-04-17
feat(editimagemodal): edit image modal to 2 pages

fysheets:
## [EDUCATOR-2672](https://openedx.atlassian.net/browse/EDUCATOR-2672)

Make the Insert/Edit image modal two pages instead of one


src/components/EditImageModal/displayMessages.jsx
```diff
@@ -96,6 +101,16 @@ const messages = defineMessages({
...
   },
+  editImageModalNextPageButton: {
+    id: 'editImageModalNextPageButton',
+    defaultMessage: 'Next',
+    description: 'Button text to go to the next screen',
+  },
+  editImageModalPreviousPageButton: {
+    id: 'editImageModalPreviousPageButton',
+    defaultMessage: 'Back',
+    description: 'Button text to go to the next screen',
```
efischer19: s/next/previous here FYI

src/components/EditImageModal/index.jsx
```diff
@@ -533,7 +643,42 @@ export default class EditImageModal extends React.Component {
...
+
+  getPreviousPageButton = () => (
+    <Button
+      label={
+        <WrappedMessage
+          message={messages.editImageModalPreviousPageButton}
+        />
+      }
+      buttonType="link"
+      onClick={this.onPreviousPageButtonClick}
```
MichaelRoytman: I think you should bind this as well.

src/components/EditImageModal/index.jsx
```diff
@@ -68,6 +80,17 @@ export default class EditImageModal extends React.Component {
...
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.pageNumber === 1 && prevState.pageNumber === 2) {
+      this.imageSourceInputRef.focus();
+    }
+
+    if (this.state.pageNumber === 2 && prevState.pageNumber === 1
+      && this.state.shouldShowPreviousButton) {
+      this.previousPageButtonRef.focus();
```
MichaelRoytman: I have come across a case where this throws an error.

To reproduce:

1. Insert an image with the modal.
2. Open the modal again to insert an image. This resets the page to 1. Do not insert a new image. Close the modal.
3. Attempt to edit the image from step 1. There will be console errors about focus, since the previousPageButtonRef is not in the DOM.

I think tacking on `this.state.shouldShowPreviousButton` to the condition should fix it.

src/components/EditImageModal/index.jsx
```diff
@@ -234,13 +269,41 @@ export default class EditImageModal extends React.Component {
   }
 
   onStatusAlertClose = () => {
-    this.imageSourceInputRef.focus();
+    if (this.state.pageNumber === 1) {
+      this.imageSourceInputRef.focus();
+    } else if (this.state.pageNumber === 2) {
+      this.previousPageButtonRef.focus();
```
MichaelRoytman: I think we should add a case where `this.state.shouldShowPreviousButton` is false (i.e. editing an existing image), since this will throw the same error as in the componentDidUpdate method. I'm not sure what we should focus on. The first fieldset?

src/components/EditImageModal/index.jsx
```diff
@@ -543,9 +688,14 @@ export default class EditImageModal extends React.Component {
       buttonType="primary"
       onClick={this.onInsertImageButtonClick}
     />
-  )
+  );
 
   handleOpenModal = (event) => {
+    let isEventSourceEmpty = true;
+    if (event.detail.src) {
+      isEventSourceEmpty = false;
```
MichaelRoytman: I can't leave a comment on these lines, but I think you could use this variable for lines 709 and 713 as well.

src/components/EditImageModal/index.jsx
fysheets: Can do!

src/components/EditImageModal/index.jsx
fysheets: This if statement already has the `shouldShowPreviousButton` check added to it, did you mean somewhere else?

src/components/EditImageModal/index.jsx
fysheets: Good catch -- I think the first Fieldset sounds good, may end up being the search bar instead once that PR merges in.

src/components/EditImageModal/index.jsx
fysheets: Cool, I'll check it out.

src/components/EditImageModal/index.jsx
MichaelRoytman: Oh, strange. I think my branch was just out of sync with your. My mistake.

src/components/EditImageModal/index.jsx
MichaelRoytman: I don't think I see this being bound? Am I missing it somewhere?

PR #184, opened 2018-04-19
Improve Automation 🤖

efischer19:
- Cleans up travis config to use build stages
- adds travisbuddy to make PR feedback faster
- adds commitlint to enforce normalized commit messages before merge to master
- add semantic-release to automatically deploy new releases on merge to master

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,0 +1,27 @@
+{
+  "branch": master",
+  "tagFormat": "v${version}",
+  "verifyConditions": [
+    "@semantic-release/npm",
+    {
+      "path": "@semantic-release/github",
+      "assets": {
+        "path": "dist/*"
+      }
+    }
+  ],
+  "analyzeCommits": "@semantic-release/commit-analyzer",
+  "generateNotes": "@semantic-release/release-notes-generator",
+  "prepare": "@semantic-release/npm",
+  "publish": [
+    "@semantic-release/npm",
+    {
+      "path": "@semantic-release/github",
+      "assets": {
+        "path": "dist/*"
+      }
+    }
+  ],
+  "success": [],
+  "fail": []
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,36 @@
 language: node_js
-
 node_js:
   - 8.9.4
 
+branches:
+  only:
+    - master
+    - /v\d+\.\d+\.d+$/
+
 cache:
-  directory:
-    - studio-frontend/node_modules/
+  directories:
+    - "node_modules"
 
-script:
-  - make package-lock.validate
-  - make i18n.pre_validate
-  - npm run lint
-  - npm run test
+stages:
+  - PR checks ✅
+  - name: Ship It 🚢
+    if: env(TRAVIS_BRANCH) = master
+jobs:
+  include:
+    - stage: PR checks ✅
+      env:
+        - NAME=Unit Tests
+      script: npm run test
+      after_script:
+        - codecov
+    - stage: PR checks ✅
+      env:
+        - NAME=Validation Checks
+      script: ./.travis/validate_changes_for_merge.sh
+    - stage: Ship It 🚢
+      env:
+        - NAME=Publish
+      script: rm -rf dist/ && rm -rf node_modules/ && npm install && npm run build && semantic-release
 
-after_script:
-  - codecov
+notifications:
+  webhooks: https://www.travisbuddy.com/
--- a/package.json
+++ b/package.json
@@ -40,6 +40,9 @@
     "whatwg-fetch": "^2.0.3"
   },
   "devDependencies": {
+    "@commitlint/cli": "^6.1.3",
+    "@commitlint/config-angular": "^6.1.3",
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -9,6 +9,7 @@
     "lint": "eslint --ext .js --ext .jsx .",
     "lint_fix": "eslint --fix --ext .js --ext .jsx .",
     "prepare": "npm run build",
+    "semantic-release": "semantic-release pre && npm publish && semantic-release post",
```
jaebradley: Since `v9.0.0` this has just been `semantic-release`

image

efischer19: Gracias, I had just copied this from your initial PR to start

package.json
```diff
@@ -78,6 +79,7 @@
     "reactifex": "1.1.1",
     "redux-mock-store": "^1.3.0",
     "sass-loader": "^6.0.7",
+    "semantic-release": "^15.1.7",
```
jaebradley: Probably want to add `travis-deploy-once` as per [the `v12.0.0` release recommendation](https://github.com/semantic-release/semantic-release/releases/tag/v12.0.0)

image

jaebradley:
@efischer19 I built https://github.com/jaebradley/semantic-release-example in case you're looking for a simplified "live" example

.travis.yml
```diff
@@ -1,17 +1,40 @@
...
 
+branches:
+  only:
+    - master
+    - /v\d+\.\d+\.d+$/
+
+before_install:
+  - npm install -g npm@latest
+install:
+  - npm ci
```
efischer19: I'm gonna revert this section; `npm ci` is only marginally faster (~75 seconds versus ~90), and the `before_install` step takes at least ~20 seconds so we don't gain anything. A regular `npm install` will be simpler.

efischer19:
In the absence of having the proper bot account credentials, I've set up https://github.com/efischer19-ci and https://www.npmjs.com/~efischer19-ci to do the publishes in the interim.

The npm account is not a member of the edx org, so it will fail when it tries to push a new release. The auth flow will get a workout though.

.travis.yml
```diff
@@ -1,17 +1,36 @@
...
+      after_script:
+        - codecov
+    - stage: PR checks ✅
+      env:
+        - NAME=Validation Checks
+      script: ./scripts/validate_changes_for_merge.sh
+    - stage: Ship It 🚢
+      env:
+        - NAME=Publish
+      script: rm -rf dist/ && rm -rf node_modules/ && npm ci && npm run build && semantic-release
```
efischer19: change to install

Makefile
```diff
@@ -83,3 +83,6 @@ package-lock.validate:
 
 copy-dist:
 	for f in dist/*; do docker cp $$f edx.devstack.studio:/edx/app/edxapp/edx-platform/node_modules/@edx/studio-frontend/dist/; done
+
+validate:
+	./scripts/validate_changes_for_merge.sh && echo "Validated, ready to merge!" || echo "Validation problems, unable to merge"
```
efischer19: this will fail locally, remove it

scripts/validate_changes_for_merge.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
efischer19: move to .travis/

package.json
```diff
@@ -40,6 +40,9 @@
     "whatwg-fetch": "^2.0.3"
   },
   "devDependencies": {
+    "@commitlint/cli": "^6.1.3",
```
efischer19: add a make command for this

.releaserc
```diff
@@ -0,0 +1,27 @@
+{
```
jaebradley: @efischer19 why specify this file? It seems like a duplicate of [the default plugins that `semantic-release`](https://github.com/semantic-release/semantic-release/blob/caribou/docs/usage/plugins.md#plugins) comes with?

.travis.yml
```diff
@@ -1,17 +1,36 @@
...
+branches:
+  only:
+    - master
+    - /v\d+\.\d+\.d+$/
+
 cache:
-  directory:
-    - studio-frontend/node_modules/
+  directories:
+    - "node_modules"
```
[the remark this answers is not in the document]
jaebradley: I don't think this will matter much at the end of the day, but I think it's slightly better to cache `~/.npm` vs. `node_modules` even though I know Travis' official docs say to use `node_modules`

See 

* [SO article](https://stackoverflow.com/a/42523517/5225575)
* [GH Issue + linked Twitter thread](https://github.com/semantic-release/cli/issues/125)

.travis.yml
```diff
@@ -1,17 +1,36 @@
...
 
-script:
-  - make package-lock.validate
-  - make i18n.pre_validate
-  - npm run lint
-  - npm run test
+stages:
+  - PR checks ✅
+  - name: Ship It 🚢
+    if: env(TRAVIS_BRANCH) = master
```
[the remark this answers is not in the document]
efischer19: supposedly the same, but it didn't actually work.

https://github.com/edx/studio-frontend/pull/189/files

PR #191, opened 2018-04-26
Use Paragon Fieldset, validate form on insert submit

thallada:
screenshot-20180426134929-1003x671

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -232,9 +232,9 @@
       }
     },
     "@edx/paragon": {
-      "version": "2.6.1",
-      "resolved": "https://registry.npmjs.org/@edx/paragon/-/paragon-2.6.1.tgz",
-      "integrity": "sha512-mVtpFUZl/xlT4NHzUM3EMtLyPBR6lkdcnrs7YSC8AOLILij0feL75vByFijL9HxXkNhidaH8XNWe+KPeJ9q81Q==",
+      "version": "2.7.0",
+      "resolved": "https://registry.npmjs.org/@edx/paragon/-/paragon-2.7.0.tgz",
+      "integrity": "sha512-YbZpn8/84ddsAsh/vWP63Zi9uRTXrvDLVP4ZJqLuQjJhts6OVetXvfWFHON/3fHK4Yvp090xaAWzbpxwJJ88Uw==",
       "requires": {
         "@edx/edx-bootstrap": "1.0.0",
         "babel-polyfill": "6.26.0",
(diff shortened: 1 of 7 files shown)


package-lock.json
```diff
@@ -6039,7 +6039,6 @@
       "resolved": "https://registry.npmjs.org/fsevents/-/fsevents-1.2.2.tgz",
       "integrity": "sha512-iownA+hC4uHFp+7gwP/y5SzaiUo7m2vpa0dhpzw8YuKtiZsz7cIXsFbXpLEeBM6WuCQyw1MH4RRe6XI8GFUctQ==",
       "dev": true,
-      "optional": true,
```
fysheets: I think the removal of this will explode things -- @efischer19 ran into this, this week.

efischer19: https://media.giphy.com/media/MVUyVpyjakkRW/giphy.gif

src/components/EditImageModal/index.jsx
```diff
@@ -551,6 +582,24 @@ export default class EditImageModal extends React.Component {
     return feedback;
   }
 
+  validateImageDimensions = () => {
+    let feedback = { isValid: true };
+    const { height, width } = this.state.imageDimensions;
+
+    if (((height !== '' && height !== undefined) && (!Number.isInteger(height) || height <= 0)) ||
+      ((width !== '' && width !== undefined) && (!Number.isInteger(width) || width <= 0))) {
```
fysheets: This if statement is really hard to read -- I thought we were only checking if these values were negative? Can we make this more readable?

package-lock.json
thallada: oh good catch. My docker container is being very insistent about removing these lines for some reason...

src/components/EditImageModal/index.jsx
thallada: Because we allow the fields to be empty, I need to check if they are `''` or `undefined` which complicates things. I'm going to abstract this condition to a separate function, what do you think of this?

```javascript
  isPositiveIntegerOrEmpty = (val) => {
    if (val === '' || val === undefined) {
      return true;
    }
    if (Number.isInteger(val) && val > 0) {
      return true;
    }
    return false;
  }
```

efischer19: 👍

fysheets: Wayy easier to read :) Thanks!

package-lock.json
efischer19: Is it possible our node version is getting updated unexpectedly? 8.1.11 came out less than a month ago according to https://nodejs.org/en/download/releases/

thallada: Hmm. Maybe I did accidentally run `npm install` on my host. Cause, I just ran `git checkout origin/master -- package-lock.json` and then `npm install` again and it's not removing the optional line anymore.

thallada:
Ok team @edx/educator-dahlia, tests are passing now so this is ready for a re-review.

PR #197, opened 2018-04-30
feat(editimagemodal): add assets dropzone to image modal

fysheets:
Add the assets dropzone upload component to the in context image modal


dsjen:
So exciting! Is there a way to limit the upload to just images? Also, I believe upon upload, the user should be taken to the second page of the modal.

fysheets:
@dsjen Good catch on limiting upload types! I'll add that to my todo list on the subtask :) (Had already captured the need to automatically move to the next page)

src/SFE.scss
```diff
@@ -34,7 +34,6 @@
   font-weight: $font-weight-base;
   line-height: $line-height-base;
   color: $body-color;
-  text-align: left;
```
thallada: Just curious, what was this style breaking?

Now that I am thinking about this style I think it might break RTL support, so it's probably better that it is left out.

src/components/AssetsDropZone/AssetsDropZone.scss
```diff
@@ -25,12 +26,16 @@
   }
 }
 
+.drop-zone-compact {
```
dsjen: Could these styles be generally applied when the screen in small? If so, media queries might be the way to go, e.g. https://github.com/edx/studio-frontend/blob/2dc352af824a6963a8b4ef905bbcfaec1407dc6e/src/components/EditImageModal/EditImageModal.scss#L69, with a mobile-first design.

fysheets: Perhaps? Do you mean to have it handle both for this instance + smaller screens? I'm not familiar enough to know if I use the media query and view the modal on my desktop if the UI will still leverage the compact styles? I'm assuming no, but happy to discuss this option more.

dsjen: I'm actually just thinking that you could combine the CSS with the above rather than having a separate selector. For the case of the modal, it would just end up being smaller. It's probably easier to show you what I mean. :)

dsjen: I just had an idea about this. You could have `drop-zone` inherit from `drop-zone-compact` and leave the media queries in `drop-zone`.

src/components/AssetsDropZone/index.jsx
```diff
@@ -79,7 +91,12 @@ export default class AssetsDropZone extends React.Component {
                   }}
                 >
                   { displayText => (
-                    <p aria-hidden id="asset-drop-zone-max-file-size-label" data-identifier="asset-drop-zone-max-file-size-label">
+                    <p
+                      className={styles['center-text']}
```
MichaelRoytman: Nit: you could alphabetize these attributes.

src/data/actions/assets.js
```diff
@@ -319,6 +319,11 @@ export const uploadExceedMaxSize = maxFileSizeMB => ({
   maxFileSizeMB,
 });
 
+export const uploadInvalidFileType = validFileTypes => ({
```
MichaelRoytman: I'm not sure I understand what validFileTypes does here. Is it intended for something later?

src/components/AssetsDropZone/index.jsx
fysheets: Easy :) can do!

src/data/actions/assets.js
fysheets: Ah I was passing this through for this logic: https://github.com/edx/studio-frontend/blob/51a300ee1f766a94e7c5eaedbc11d52eae8f7b8b/src/components/AssetsDropZone/index.jsx#L27 

`else if (rejectedFiles.length > 0) {
      // bubbling up error from the first rejected file in the list
      const rejectedFile = rejectedFiles[0];
      if (rejectedFile.size > this.dropZoneMaxFileSizeBytes) {
        this.props.uploadExceedMaxSize(this.dropZoneMaxFileSizeBytes);
      } else {
        this.props.uploadInvalidFileType(this.props.acceptedFileTypes);
      }
    }`

The `else` logic. I was hoping to make that an `else if` to be more specific, but I think to do that effectively we'll need to import another package (as the dropzone uses an external package for that logic) and I ended up with the plain else instead. I'll go through and clean this up for now as well. If we decide to make that more explicit we can.

src/components/EditImageModal/EditImageModal.scss
```diff
@@ -1,7 +1,7 @@
 @import 'edx-bootstrap';
 
 .modal-body {
-  height: 400px;
+  height: 75%;
```
thallada: I wonder if we could set a max-width instead here. I think the modal will get really smooshed if the window height is scaled way down.

src/components/AssetsDropZone/AssetsDropZone.scss
```diff
@@ -4,19 +4,21 @@
   text-align: center;
 }
 
-.drop-zone {
+.drop-zone-compact {
   outline: $dark dashed 1px;
-  padding: $spacer;
+  padding: 0;
```
thallada: I think we could use a *touch* more padding here because the dotted line is riding up right next to the icon and max file size text. Maybe `$spacer / 4`?

PR #198, opened 2018-05-01
feat(editimagemodal): mroytman/assets list

MichaelRoytman:


src/components/AssetsList/AssetsList.test.jsx
```diff
@@ -0,0 +1,608 @@
...
+
+      resetWrapper();
+    });
+
+    it('this.state.selectedAssetPage is set by clicking on a listbox option', () => {
+      items.at(1).simulate('mouseDown');
+
+      expect(wrapper.state('selectedAssetPage')).toEqual(0);
+
+      resetWrapper();
```
thallada: I think these `resetWrapper()` calls could go into a `beforeEach` or `afterEach` function for this `describe` block.

src/components/AssetsList/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+        ...initialState,
+      });
+    }
+  }
+
+  onListBoxFocus = () => {
+    // if no list item is selected, select first item on  listbox focus
+    if (this.state.selectedAssetIndex === -1) {
+      this.setState(
+        { selectedAssetIndex: this.state.selectedAssetIndex + 1,
```
thallada: Why not just pass `0` here? That way this `setState` call won't have to rely on existing state.

src/components/AssetsList/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+  );
+
+  getDateAddedElement = (dateAdded, index) => (
+    <span id={`asset-date-${index}`} className="col-3" data-identifier="asset-date-added">{dateAdded}</span>
+  );
+
+  getAssetListItem = (asset, index) => {
+    const isSelected = this.props.selectedAsset.id === asset.id;
+
+    return (<li
```
thallada: Nit: I think it would look a little better if this `<li` started on a newline below.

src/components/AssetsList/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+
+  handleKeyDown = (e) => {
+    switch (e.key) {
+      case 'ArrowDown': {
+        // prevent scrolling entire modal body with arrow keys
+        e.preventDefault();
+
+        if (this.state.selectedAssetIndex < this.props.assetsListMetadata.pageSize - 1) {
+          this.setState(
+            { selectedAssetIndex: this.state.selectedAssetIndex + 1,
```
thallada: Because this `setState` relies on existing state I think we need to use the [functional `setState`](https://medium.freecodecamp.org/functional-setstate-is-the-future-of-react-374f30401b6b).

```jsx
this.setState((state, props) => ({
    selectedAssetIndex: state.selectedAssetIndex + 1,
    selectedAssetPage: props.assetListMetadata.page,
}, () => {
    this.props.selectAsset(this.props.assetsList[this.state.selectedAssetIndex]);
}));
```

src/components/AssetsList/index.jsx
```diff
@@ -0,0 +1,221 @@
...
+        }
+        break;
+      }
+      case 'ArrowUp': {
+        // prevent scrolling entire modal body with arrow keys
+        e.preventDefault();
+
+        if (this.state.selectedAssetIndex > 0) {
+          this.setState(
+            { selectedAssetIndex: this.state.selectedAssetIndex - 1,
```
thallada: Same thing here with functional setState.

src/components/EditImageModal/index.jsx
```diff
@@ -45,47 +43,48 @@ export default class EditImageModal extends React.Component {
...
+      this.setState({
+        imageSource: nextProps.selectedAsset.portable_url,
+      });
+    }
+  }
+
   componentDidUpdate(prevProps, prevState) {
     if (this.state.pageNumber === 1 && prevState.pageNumber === 2) {
-      this.imageSourceInputRef.focus();
+      // TODO: add focus
```
thallada: Is this supposed to focus on the drop zone here?

src/utils/scss/focus.scss
```diff
@@ -0,0 +1,8 @@
+@mixin focus() {
+    @include form-control-focus();
+    border-width: 2px;
+    outline: none;
+    box-shadow: $input-box-shadow, $input-btn-focus-box-shadow;
+}
+
+
```
thallada: Nit: extra lines and whitespace at the end here.

src/components/EditImageModal/index.jsx
MichaelRoytman: Maybe, but I don't have the drop zone on my branch, which is why I'm leaving this here. Does that seem okay?

thallada: Oh yeah :+1:, I just wanted to make sure.

PR #205, opened 2018-05-03
fix(assets): give copy urls header more width

thallada:
This prevents the header text from wrapping and making the header row taller.

I changed it from affecting all `th` elements to only the "Copy URLs" one because the old rule was making the header row extra long and forcing the table to always overflow and have a vertical scrollbar.

### DIFF
--- a/src/components/AssetsTable/AssetsTable.scss
+++ b/src/components/AssetsTable/AssetsTable.scss
@@ -43,11 +43,15 @@
 }
 
 .table thead th {
-  min-width: 12rem;
-
   button {
     padding: 0;
   }
+}
+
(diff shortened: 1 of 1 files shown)


fysheets:
I don't love the specific edge case-y-ness of this. Could you add a comment on why it's the `nth-child(5)`? That way when we come back to this in a few weeks/months/etc someone can understand what that's for. Or what does this look like without that min-width setting at all? Does that one column _really_ need that setting if the rest of them don't?

thallada:
I don't like having to use `:nth-child` either but the Paragon table
doesn't support adding a one-off class to one specific column header. I
think it will be easier to show you visually what's going on tomorrow and
why this style fixes it.

PR #210, opened 2018-05-10
make EditImageModal tests work and convert CopyButton tests to shallow

MichaelRoytman:


MichaelRoytman:
I have tried to summarize some findings I have made during the course of fixing our EditImageModal tests and converting our use of mount to shallow.

**1. Due to https://github.com/airbnb/enzyme/issues/664, context is not passed down to the root wrapper's children when using shallow.**

For example, given

```
const wrapper = shallowWithIntl(
  <Provider store={store}>
    <WrappedEditImageModal />
  </Provider>,
  { disableLifecycleMethods: true },
);
```

we would expect that any of the wrapper's children will be passed the store as part of the parent's context. We would expect to be able to do the following.

`const editImageModal = wrapper.find('Connect(EditImageModal)').dive();`

However, due to the aforementioned bug, the Connect(EditImageModal) (i.e. WrappedEditImageModal) component complains about not being passed a store. Therefore, we need to explicitly pass the necessary context to a component on every dive. That is, the following

`const editImageModal = wrapper.find('Connect(EditImageModal)').dive({ context: { store, intl } });`

Assume that store and intl are already defined. 

I have tried to abstract some of the commonly needed components into functions to avoid the visual complexity of such messy dives in every test case.

**2. The state of a wrapper is not persisted across dives();.**

``` 
editImageModal = (wrapper.find('Connect(EditImageModal)').dive({ context: { store, intl } }).find(EditImageModal).dive({ context: { store, intl } }));
editImageModal.setState({ open: true });
console.log(editImageModal.state('open'));

editImageModal = (wrapper.find('Connect(EditImageModal)').dive({ context: { store, intl } }).find(EditImageModal).dive({ context: { store, intl } }));
console.log(editImageModal.state('open'));
```

Here, we get the open state is true at the first log statement, but the open state is false for the second log statement. This is because the wrapper is a reference to Connect(EditImageModal), which is the connected component. This component is not the component that has state. It is just a higher order component that wraps the actual component and connects it to the Redux store. Therefore, a change in the state of its child does not modify it. By grabbing a fresh reference to the EditImageModal in the third line, we get a reference that does not reflect the state change made in the second line, since the wrapper it dives on is agnostic to such state changes. Of course, an individual reference will maintain changes you make.

**3. If you make a change to a component that will affect the rendering of its children, you may need to call the update function on the root component before asserting on the props, state, or rendering of the children. You may need to do this update before getting a fresh reference to the child, as references set before the update function is called may be stale.**

For example, see the following.

```
editImageModal.setState({
  imageDimensions: {
    height: -100,
    width: -100,
  },
  pageNumber: 2,
  shouldShowPreviousButton: true,
});

const insertImageButton = getInsertImageButton(editImageModal);
insertImageButton.simulate('click');

editImageModal.update();

...

const imageDimensionsFieldset = getImageDimensionsFieldset(editImageModal);
expect(imageDimensionsFieldset.prop('isValid')).toEqual(false);
expect(imageDimensionsFieldset.prop('invalidMessage').props.message).toEqual(messages.editImageModalFormValidImageDimensions);
```

To be honest, I don't completely understand how this works, especially with shallow. It's a little more clear with mount. You can read about mount and update [here](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#for-mount-updates-are-sometimes-required-when-they-werent-before). The following is just my hunch, but I have not been able to find a comprehensive explanation of why update is sometimes necessary. The above test starts off by setting the imageDimensions to be invalid values. It then clicks the "Insert Image" button, expecting that the invalid image dimensions will trigger a change in the EditImageModal's state to reflect this invalidity, cascading down to the imageDimensionsFieldset, which should receive new props reflecting this invalidity. However, the act of clicking on the "Insert Image" button does not cause the EditImageModal children to re-render, which includes the imageDimensionsFieldset. It only re-renders the EditImageModal itself. Therefore, the props to the imageDimensionsFieldset have not actually changed, even though the EditImageModal state that controls those props (`isImageDimensionsValid`) has changed appropriately. Calling editImageModal.update() forces the EditImageModal to re-render, which, somehow, pushes these props down through to the children. The test can then correctly assert on the props of the imageDimensionsFieldset. I am not sure why the re-render that occurs as part of the button click does not have the same behavior.

**4. shallow does not handle refs. This can cause problems.**

If you refer to refs in either the componentDidMount or the componentDidUpdate lifecycle method, your test may throw ref related errors before you have even been able to shallow a component. This could be because componentDidMount runs immediately on shallowing a component or because something is triggering a componentDidUpdate immediately. In either case, you can pass the disableLifeCycleMethods option to shallow. ["If set to true, componentDidMount is not called on the component, and componentDidUpdate is not called after setProps and setContext."](http://airbnb.io/enzyme/docs/api/shallow.html#arguments) It defaults to false. Documentation about this addition to Enzyme 3 are [here](http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#lifecycle-methods).

If your test throws ref related errors during execution, you can mock the ref before performing the action that triggers an interaction with a ref.

For example, see the following click handler for the "Insert Image" button.

```
onInsertImageButtonClick = () => {
  const isValidImageDescription = this.validateImageDescription();
  const isValidImageDimensions = this.validateImageDimensions();
  const isValidFormContent = isValidImageDescription.isValid && isValidImageDimensions.isValid;

  const currentValidationMessages = {};

  if (!isValidImageDescription.isValid) {
    currentValidationMessages[imageDescriptionID] = isValidImageDescription.validationMessage;
  }

  if (!isValidImageDimensions.isValid) {
    currentValidationMessages[imageWidthID] = isValidImageDimensions.validationMessage;
  }

  this.setState({
    isStatusAlertOpen: !isValidFormContent,
    isImageDescriptionValid: isValidImageDescription.isValid,
    isImageDimensionsValid: isValidImageDimensions.isValid,
    currentValidationMessages,
  });

  if (isValidFormContent) {
    this.imageFormRef.dispatchEvent(new CustomEvent('submitForm',
    {
       bubbles: true,
       detail: {
         height: this.getDimensionStateOrNatural('height'),
         width: this.getDimensionStateOrNatural('width'),
         src: this.state.imageSource,
         alt: this.state.isImageDecorative ? '' : this.state.imageDescription,
         style: this.state.imageStyle,
       },
    },
    ));

    this.setState({
      open: false,
    });

    this.resetImageSelection();
  } else {
    this.statusAlertRef.focus();
  }
}
```

Notice that when clicking the "Insert Image" button, if a form is determined to be invalid, we focus on the statusAlertRef. However, because shallow does not support refs, when clicking the "Insert Image" button from a test, the test throws an error, since statusAlertRef is not defined.

For example see the following test.

```
editImageModal.setState({
  imageDimensions: {
    height: -100,
    width: -100,
  },
});

const insertImageButton = getInsertImageButton(editImageModal);
insertImageButton.simulate('click');
```
Since the image dimensions are invalid, the EditImageModal attempts to focus on the statusAlertRef.

What we can do is mock the ref in the following way before clicking on the "Insert Image" button.

```
const focusSpy = jest.fn();

const statusAlertRefMock = {
  focus: focusSpy,
};

editImageModal.instance().statusAlertRef = statusAlertRefMock;
```

Notice that we have to mock the focus function as well, not just the ref, since a object does not have a focus function by default. You do not need to pass in a Jest mock function if you do not need it. Perhaps you just need the EditImageModal to not choke on that line and you do not care how many times or with what parameters focus is called. 

In that case, either of these will work.

```
const statusAlertRefMock = {
  focus: () => {},
};
```
```
const statusAlertRefMock = {
  focus: jest.fn(),
};
```

However, passing in a Jest mock function that you keep a reference to allows you to assert on the use of that function.

MichaelRoytman:
Oh, I also converted the CopyButton tests to use shallow to get some practice before dealing with the EditImageModal, so that's why that test file has been changed as well.

fysheets:
One thing I just realized - I know we had removed the EditImageModal test file from coverage, we should re-add it with this PR as well.

MichaelRoytman:
@fysheets I believe I did in the package.json file. That's why the coverage dropped a little. Does this look okay to merge?

fysheets:
🚢

PR #211, opened 2018-05-11
fix(paragon): upgrade to paragon 3.0.5

thallada:
Fix broken models (EditImageModal and the delete asset modal on the AssetsPage) after breaking change update. Basically, the `Modal`s are now wrapped in a div with an `id` and the `Modal`'s `parentSelector` prop is set to that `id` so that the modal is rendered in the same place in the DOM that it appears in the JSX.

The `EditImageModal` sends a signal when the modal is closed now too. My upcoming platform PR will have the Studio JS make the background body un-scrollable when the modal is opened and then listen for this signal and make it scrollable again when the modal closes.

Moved the modal Sass from component Sass files to the common `SFE.scss` that is shared by all components.

"Why are we copying Sass from Paragon into our Sass files? That seems dumb." Yeah it is, we have [a ticket (EDUCATOR-2704)](https://openedx.atlassian.net/browse/EDUCATOR-2704) for figuring out how to import the Paragon CSS instead. The problem with just doing that now is that the Paragon's themeable CSS includes Bootstrap and FontAwesome, and we only want just the Paragon-specific styles.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^1.0.0",
-    "@edx/paragon": "2.7.0",
+    "@edx/paragon": "3.0.5",
     "babel-polyfill": "^6.26.0",
     "classnames": "^2.2.5",
     "copy-to-clipboard": "^3.0.8",
@@ -32,7 +32,6 @@
     "react-paginate": "^5.0.0",
     "react-redux": "^5.0.6",
     "react-transition-group": "^2.2.1",
-    "reactstrap": "^4.8.0",
     "redux": "^3.7.2",
     "redux-devtools-extension": "^2.13.2",
     "redux-thunk": "^2.2.0",
@@ -108,7 +107,9 @@
       "<rootDir>/src/components/BackendStatusBanner/",
       "<rootDir>/src/.*/container.jsx",
       "<rootDir>/src/data/i18n/locales/",
-      "<rootDir>/src/utils/matches-prefixer.js"
+      "<rootDir>/src/utils/matches-prefixer.js",
+      "<rootDir>/src/utils/mockPortals.js",
+      "<rootDir>/src/utils/mockQuerySelector.js"
     ],
     "transformIgnorePatterns": [
       "/node_modules/(?!(@edx/paragon)/).*/"
--- a/src/components/AssetsTable/AssetsTable.test.jsx
+++ b/src/components/AssetsTable/AssetsTable.test.jsx
@@ -4,6 +4,7 @@ import AssetsTable from './index';
 import courseDetails, { testAssetsList } from '../../utils/testConstants';
 import { assetActions } from '../../data/constants/actionTypes';
 import { assetLoading } from '../../data/constants/loadingTypes';
+import mockQuerySelector from '../../utils/mockQuerySelector';
 import { mountWithIntl } from '../../utils/i18n/enzymeHelper';
 import AssetsStatusAlert from '../AssetsStatusAlert/index';
 
@@ -84,6 +85,12 @@ const numberOfWebButtons = defaultProps.assetsList.filter(asset => asset.externa
 let wrapper;
 
 describe('<AssetsTable />', () => {
+  beforeEach(() => {
+    mockQuerySelector.init();
+  });
+  afterEach(() => {
+    mockQuerySelector.reset();
+  });
   describe('renders', () => {
     beforeEach(() => {
       wrapper = mountWithIntl(
@@ -301,7 +308,7 @@ describe('<AssetsTable />', () => {
       trashButtons.at(0).simulate('click');
 
       modal = wrapper.find('[role="dialog"]');
-      expect(modal.hasClass('modal-open')).toEqual(true);
+      expect(modal.hasClass('show')).toEqual(true);
       expect(wrapper.state('modalOpen')).toEqual(true);
     });
     it('closes when Cancel button clicked; modalOpen state is false', () => {
@@ -463,11 +470,17 @@ describe('Lock Asset', () => {
   const getLockingButtons = () => wrapper.find('button > .fa-spinner');
 
   beforeEach(() => {
+    mockQuerySelector.init();
+
     wrapper = mountWithIntl(
       <AssetsTable
         {...defaultProps}
       />,
     );
+  });
+
+  afterEach(() => {
+    mockQuerySelector.reset();
   });
 
   it('renders icons and buttons', () => {
--- a/src/components/EditImageModal/EditImageModal.test.jsx
+++ b/src/components/EditImageModal/EditImageModal.test.jsx
@@ -6,6 +6,8 @@ import React from 'react';
 import EditImageModal from './index';
 import { getMockStore } from '../../utils/testConstants';
 import messages from './displayMessages';
+import mockModalPortal from '../../utils/mockModalPortal';
+import mockQuerySelector from '../../utils/mockQuerySelector';
 import rewriteStaticLinks from '../../utils/rewriteStaticLinks';
 import { shallowWithIntl } from '../../utils/i18n/enzymeHelper';
 import WrappedEditImageModal from './container';
@@ -72,7 +74,15 @@ let editImageModal;
 
 describe('EditImageModal', () => {
   beforeEach(() => {
+    mockQuerySelector.init();
+    mockModalPortal.init();
+
     editImageModal = getEditImageModal(wrapper);
+  });
+
+  afterEach(() => {
+    mockModalPortal.reset();
+    mockQuerySelector.reset();
   });
 
   describe('page 1 renders', () => {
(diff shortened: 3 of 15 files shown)


package.json
```diff
@@ -32,7 +32,6 @@
     "react-paginate": "^5.0.0",
     "react-redux": "^5.0.6",
     "react-transition-group": "^2.2.1",
-    "reactstrap": "^4.8.0",
```
thallada: I think this was left here by mistake at some point. We don't use reactstrap anywhere in SFE.

thallada:
Got a pretty weird test failure here. I think it has something to do with having to render the modal on the AssetsPage in a function so that we can wrap the `closeText` string in a `<span lang>`.

I'm hoping that [this PR in Paragon that will allow passing an element to `closeText`](https://github.com/edx/paragon/pull/245) will simplify the `Modal` rendering and get rid of the error (since the `EditImageModal` tests passed fine).

src/components/AssetsTable/AssetsTable.test.jsx
```diff
@@ -81,9 +81,21 @@ const numberOfImages = defaultProps.assetsList.filter(asset => asset.thumbnail).
...
 const numberOfWebButtons = defaultProps.assetsList.filter(asset => asset.external_url).length;
 
+let parentElement;
+let querySelector;
 let wrapper;
 
 describe('<AssetsTable />', () => {
+  beforeEach(() => {
+    querySelector = global.document.querySelector;
+    parentElement = global.document.createElement('div');
```
MichaelRoytman: What does `global` do? I thought that `document.createElement('div')` would work.

src/components/AssetsTable/AssetsTable.test.jsx
```diff
@@ -81,9 +81,21 @@ const numberOfImages = defaultProps.assetsList.filter(asset => asset.thumbnail).
 const numberOfStudioButtons = defaultProps.assetsList.filter(asset => asset.portable_url).length;
 const numberOfWebButtons = defaultProps.assetsList.filter(asset => asset.external_url).length;
 
+let parentElement;
+let querySelector;
 let wrapper;
 
 describe('<AssetsTable />', () => {
+  beforeEach(() => {
+    querySelector = global.document.querySelector;
```
MichaelRoytman: Why do you need a reference to the querySelector method? I see you're changing its implementation below, right? Does this reference get used in some way?

src/components/AssetsTable/AssetsTable.test.jsx
```diff
@@ -81,9 +81,21 @@ const numberOfImages = defaultProps.assetsList.filter(asset => asset.thumbnail).
 const numberOfStudioButtons = defaultProps.assetsList.filter(asset => asset.portable_url).length;
 const numberOfWebButtons = defaultProps.assetsList.filter(asset => asset.external_url).length;
 
+let parentElement;
+let querySelector;
 let wrapper;
 
 describe('<AssetsTable />', () => {
+  beforeEach(() => {
```
MichaelRoytman: It would be great if you would either move the contents of the beforeEach and afterEach into functions with descriptive names or leave a comment explaining what this is for. I can see it being confusing in a few months from now with fresh eyes.

src/utils/mockPortals.js
```diff
@@ -0,0 +1,41 @@
...
+const portalOrigin = {};
+
+const mockPortal = {
+  init: () => {
+    portalOrigin.render = Modal.prototype.render;
+    Modal.prototype.render = function () {
+      return this.renderModal();
+    };
+
+    portalOrigin.getMountNode = Modal.prototype.getMountNode;
```
MichaelRoytman: What is getMountNode? I don't see getMountNode in the Modal.

src/utils/i18n/enzymeHelper.jsx
```diff
@@ -32,10 +32,11 @@ export function shallowWithIntl(node, { context, disableLifecycleMethods } = {})
   );
 }
 
-export function mountWithIntl(node, { context, childContextTypes } = {}) {
+export function mountWithIntl(node, { attachTo, context, childContextTypes } = {}) {
```
MichaelRoytman: What does attachTo do? This is the only place I see it referenced, and I don't see any additional parameters to the use of mountWithIntl in the codebase. Is it still needed?

src/components/EditImageModal/index.jsx
```diff
@@ -147,6 +148,11 @@ export default class EditImageModal extends React.Component {
 
     this.props.clearSearch(this.props.courseDetails);
     this.resetImageSelection();
+    this.modalWrapperRef.querySelector('.modal').dispatchEvent(new CustomEvent('closeModal',
```
MichaelRoytman: I assume this is related to your comment, `The EditImageModal sends a signal when the modal is closed now too.`

This is solely to handle the scrolling issue in Studio, right?

Is there a reason you have to query for `.modal` instead of dispatching the event on the modalWrapperRef itself? That's how we add the eventListener to the Modal.

src/components/EditImageModal/EditImageModal.test.jsx
```diff
@@ -72,23 +76,37 @@ let editImageModal;
 
 describe('EditImageModal', () => {
   beforeEach(() => {
+    querySelector = global.document.querySelector;
+    parentElement = global.document.createElement('div');
+    global.document.body.appendChild(parentElement);
+    global.document.querySelector = () => parentElement;
+    mockPortal.init();
```
MichaelRoytman: Why does the EditImageModal test need the mockPortal.init(), but the AssetsTable one does not?

src/components/AssetsTable/AssetsTable.test.jsx
thallada: I thought global was necessary, but I removed it and it still works, so I'll do that instead.

src/components/AssetsTable/AssetsTable.test.jsx
thallada: I save a reference so I can return the function back to normal in the `afterEach` function with `document.querySelector = querySelector`.

src/utils/mockPortals.js
thallada: Oh oops. I copied this code from another project and I just assumed this was a React thing, but it's actually specific to that other project. I'll remove this.

src/utils/i18n/enzymeHelper.jsx
thallada: I briefly tried using the [attachTo](http://airbnb.io/enzyme/docs/api/ReactWrapper/detach.html) option in Enzyme, but this function wasn't passing that option through to Enzyme's `mount` function without this change. I decided to keep this here even though I stopped using `attachTo` in case we ever needed to use it in the future (otherwise we might have to go through the same confusion about why the option isn't going through).

src/components/EditImageModal/index.jsx
thallada: Yep, that's right. The `edit.js` code in the platform already had a reference to the `.modal` that it was listening to, so I made sure to emit the event where it is listening. Though, now looking at the platform code again, I think I can change it so that it is listening to the modal wrapper, and then I could dispatch this event from `this.modalWrapperRef` instead which would be cleaner.

src/components/EditImageModal/EditImageModal.test.jsx
thallada: Enzyme actually supports portals when using `mount()` now, it just doesn't work with `shallow()`. And, `AssetsTable.test.jsx` uses `mount()`.

src/components/AssetsTable/AssetsTable.test.jsx
MichaelRoytman: Oh, I see. I missed that. Thank you for explaining.

src/components/EditImageModal/index.jsx
MichaelRoytman: Ah, okay. That sounds good. I think that makes a little more sense conceptually too.

src/utils/i18n/enzymeHelper.jsx
```diff
@@ -22,20 +22,21 @@ function nodeWithIntlProp(node) {
...
     },
   );
 }
 
-export function mountWithIntl(node, { context, childContextTypes } = {}) {
+export function mountWithIntl(node, { context, childContextTypes, ...otherOptions } = {}) {
   return mount(
     nodeWithIntlProp(node),
     {
+      ...otherOptions,
```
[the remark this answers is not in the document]
MichaelRoytman: That looks good to me. Thanks!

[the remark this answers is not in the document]
MichaelRoytman: Do you think it's less clear this way? I'm fine with either option, I think, but at least this would prevent us from having to touch this function as much to add on parameters.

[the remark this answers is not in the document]
thallada: I prefer this way. Now it behaves like an actual wrapper around mount/shallow and passes through ALL options, which is what I originally expected.

PR #213, opened 2018-05-16
add EditImageModal tests

MichaelRoytman:


src/components/AssetsClearFiltersButton/AssetsClearFiltersButton.test.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { Button } from '@edx/paragon';
 
 import AssetsClearFiltersButton from './index';
-import courseDetails from '../../utils/testConstants';
+import { courseDetails } from '../../utils/testConstants';
```
MichaelRoytman: All the following courseDetails changes are because I made courseDetails a named export of testConstants instead of the default export, since I didn't think it made sense for it to be the default export.

src/utils/testConstants.jsx
```diff
@@ -5,7 +5,7 @@ import thunkMiddleware from 'redux-thunk';
 import accessibility from '../data/reducers/accessibility';
 import { assets, metadata } from '../data/reducers/assets';
 
-const courseDetails = {
+export const courseDetails = {
```
MichaelRoytman: I didn't think it made sense for this to be a default export instead of just a named export like the rest of the test constants.

src/components/EditImageModal/EditImageModal.test.jsx
```diff
@@ -85,319 +90,707 @@ describe('EditImageModal', () => {
...
-        expect(modal.find('.modal.show')).toHaveLength(0);
-        expect(modal.find('.modal-backdrop.show')).toHaveLength(0);
-      });
+  describe('renders', () => {
+    describe('page 1', () => {
+      describe('a modal with', () => {
+        it('correct initial props', () => {
+          const modal = getModal(editImageModal);
+          expect(modal.prop('open')).toEqual(false);
+          expect(shallowWithIntl(modal.prop('title')).find(WrappedMessage).prop('message')).toEqual(messages.editImageModalInsertTitle);
```
MichaelRoytman: Using shallowWithIntl here lets me shallow render the actual title prop (which is a React element) being passed to the modal, so I can assert on that wrapper instead of fishing deeper into the modal, which is Paragon's domain. All I want to test is that the prop is correct. I use this pattern a number of places where we passed React elements as props.

src/components/EditImageModal/EditImageModal.test.jsx
```diff
@@ -85,319 +90,707 @@ describe('EditImageModal', () => {
...
+          expect(shallowWithIntl(modal.prop('closeText')).prop('message')).toEqual(messages.editImageModalCancelButton);
+        });
 
-      it('an open modal when this.state.open is true', () => {
-        editImageModal.setState({ open: true });
-        const modal = getModal(editImageModal);
+        describe('buttons', () => {
+          Object.values(pageTypes).forEach((type) => {
```
MichaelRoytman: This checks that the buttons render for each page view.

fysheets:
Looks like `package-lock` updated again without a `package.json` update, but it didn't fail validation this time - once we confirm that's fine I'll swap my review.

fysheets:
Talked through it with @efischer19 and we didn't want to hold this up for the package-lock, but maybe a quick 2 minute discussion during stand-up as an FYI that this happened would be good tomorrow.

PR #215, opened 2018-05-16
fix(everything): use newer base image

efischer19:
What I did:

1. `docker-destroy-all.sh` (see below, it kills everything ded)
1. deleted `studio-frontend` directory
1. `git clone` studio frontend
1. modify `Dockerfile`, as seen in this changeset
1. `docker build -t edxops/studio-frontend:latest --no-cache .`
1. `make down`, `docker volume ls`, and `docker volume rm studio-frontend_container_only_node_modules`
1. (at this point, `package.json` had the changes you see here; `package-lock` had a small [~40ish lines] diff, mostly dropping `optional` in a few places.)
1. (on my host machine) `rm -rf node_modules/` and `rm package-lock.json`
1. `make up-detached` (which finished way too quickly), followed by `make shell` and (in container) `npm install`
1. exit shell, and commit everything you n0w see here

```
ericfischer@dhcp-18-189-45-131:~/scripts$ cat docker_destroy_all.sh 
#!/bin/bash
# Stop all containers
docker stop $(docker ps -a -q)
# Delete all containers
docker rm $(docker ps -a -q)
# Delete all images
docker rmi $(docker images -q)
# Delete all volumes
docker volume rm -f $(docker volume ls -q)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:8.9.4
+FROM node:8.11
 
 USER root
 RUN apt-get update
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,33 +5,33 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0-beta.46",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.46.tgz",
-      "integrity": "sha512-7BKRkmYaPZm3Yff5HGZJKCz7RqZ5jUjknsXT6Gz5YKG23J3uq9hAj0epncCB0rlqmnZ8Q+UUpQB2tCR5mT37vw==",
+      "version": "7.0.0-beta.47",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0-beta.47.tgz",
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.9.4
+FROM node:8.11
```
efischer19: I haven't pushed this to dockerhub yet, but I easily could

efischer19:
My working theory is that we're seeing issues because our dockerhub image is old; `library/node` [doesn't list 8.9.4 any longer](https://hub.docker.com/_/node/)

package-lock.json
```diff
@@ -5,41 +5,41 @@
...
+        "@commitlint/lint": "6.2.0",
         "@commitlint/load": "6.1.3",
         "@commitlint/read": "6.1.3",
         "babel-polyfill": "6.26.0",
         "chalk": "2.3.1",
         "get-stdin": "5.0.1",
         "lodash.merge": "4.6.1",
         "lodash.pick": "4.4.0",
-        "meow": "3.7.0"
+        "meow": "4.0.0"
```
thallada: :cat2:

fysheets:
I know that we've been refreshing our local setups on an ad hoc basis individually, but should we find some form of synchronization for this?

efischer19:
I just pushed the newly built image up to https://hub.docker.com/r/edxops/studio-frontend/

PR #219, opened 2018-05-22
fix(ci): use npm ci on travis

efischer19:
Let's give this a shot and see what happens

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,16 @@
 language: node_js
 node_js:
 - 8.11
+install:
+  - npm i -g npm@5.8.0
+  - npm ci
 branches:
   only:
   - master
   - "/v\\d+\\.\\d+\\.d+$/"
 cache:
   directories:
-  - node_modules
+  - "$HOME/.npm"
 stages:
 - PR checks ✅
 - name: "Ship It \U0001F6A2"
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ FROM node:8.11
 USER root
 RUN apt-get update
 RUN apt-get install -y vim
+RUN npm i -g npm@5.8.0
 
 WORKDIR /prebuilt
 COPY package.json .
--- a/.travis/validate_changes_for_merge.sh
+++ b/.travis/validate_changes_for_merge.sh
@@ -4,7 +4,6 @@
 
 EXIT_STATUS=0
 npm run lint || EXIT_STATUS=$?
-make package-lock.validate || EXIT_STATUS=$?
 make i18n.pre_validate || EXIT_STATUS=$?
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -1,13 +1,16 @@
 language: node_js
 node_js:
 - 8.11
+install:
+  - npm i -g npm@5.7.1 # this should be deleted once 5.7.X is out of "pre-release"
```
thallada: Any reason we shouldn't just go straight to the latest release 6.0.1? I see that fixes to `npm ci` were mentioned in the [`5.8.0` release notes](https://github.com/npm/npm/releases/tag/v5.8.0). I know that the LTS version that ships with our node version is `5.6.0` but hopefully we can avoid more NPM weirdness if we stick to the later versions where the dust has settled on the `package-lock.json` fiasco.

.travis.yml
```diff
@@ -1,13 +1,16 @@
...
+  - npm ci             # instead of default npm install
 branches:
   only:
   - master
   - "/v\\d+\\.\\d+\\.d+$/"
 cache:
   directories:
-  - node_modules
+  - "$HOME/.npm"
```
thallada: I'm assuming the node_modules folder is inside the `~/.npm/` folder?

efischer19: Yeah, and just caching `node_modules` gains us nothing when using ci, [per the docs](https://docs.npmjs.com/cli/ci):

`If a node_modules is already present, it will be automatically removed before npm ci begins its install.`

thallada:
Also, we don't need the package-lock validation check anymore with this change right? I don't think `npm ci` should ever edit the `package-lock.json` file (it actually treats it like a lock...). And if someone updates a package in `package.json` but forgets to include `package-lock.json` changes: "If dependencies in the package lock do not match those in package.json, npm ci will exit with an error, instead of updating the package lock." (https://docs.npmjs.com/cli/ci#description)

.travis.yml
thallada: If this is caching the `node_modules` folder, I'm not sure it's worth it. I just noticed this comment on the NPM docs page: "If a node_modules is already present, it will be automatically removed before npm ci begins its install." (https://docs.npmjs.com/cli/ci#description)

thallada: Oh, nevermind, I just looked it up. I guess the `~/.npm` folder is a cache, not the actual `node_modules` folder. So, I think this makes sense :+1:.

Dockerfile
```diff
@@ -3,6 +3,7 @@ FROM node:8.11
 USER root
 RUN apt-get update
 RUN apt-get install -y vim
+RUN npm i -g npm@5.8.0
```
efischer19: I'll push an update to Dockerhub for this change after lunch

PR #229, opened 2018-06-29
Add links to Course Launch Checklist

MichaelRoytman:


src/components/CourseChecklist/CourseChecklist.scss
```diff
@@ -1,10 +1,16 @@
...
+@import "~bootstrap/scss/utilities/_screenreaders";
 
-$dark-color: theme-color("dark");
+$light-color: theme-color("light");
 $lightest-color: theme-color("lightest");
 $primary-color: theme-color("primary");
 $success-color: theme-color("success");
 
+.checklist-item-link {
+  text-decoration: none;
```
MichaelRoytman: This removes the underline of the `a` element rendered by the `Hyperlink` element.

src/components/CourseChecklist/CourseChecklist.scss
```diff
@@ -31,14 +37,14 @@ $success-color: theme-color("success");
...
 
-.font-large {
-  font-size: $font-size-lg;
-}
-
 .font-small {
   font-size: $font-size-sm;
 }
 
+.font-large {
```
MichaelRoytman: I wonder if these font sizes should be put in `SFE.scss`. Any chance they'd be useful to reuse going forward?

src/components/CourseChecklist/CourseChecklist.test.jsx
```diff
@@ -1,13 +1,17 @@
...
 import getFilteredChecklist from '../../utils/CourseChecklist/getFilteredChecklist';
 import getValidatedValue from '../../utils/CourseChecklist/getValidatedValue';
+import messages from './displayMessages';
 import { shallowWithIntl } from '../../utils/i18n/enzymeHelper';
+import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
+
 
 // generating test checklist to avoid relying on actual data
-const testChecklistData = ['a', 'b', 'c', 'd'].reduce(((accumulator, currentValue) => { accumulator.push({ id: currentValue, shortDescription: currentValue, longDescription: currentValue }); return accumulator; }), []);
+const testChecklistData = ['welcomeMessage', 'gradingPolicy', 'certificate', 'courseDates', 'assignmentDeadlines'].reduce(((accumulator, currentValue) => { accumulator.push({ id: currentValue }); return accumulator; }), []);
```
MichaelRoytman: I had to use actual valid keys here because the CourseChecklist component only successfully renders if the data includes valid keys.

src/components/CourseChecklist/index.jsx
```diff
@@ -1,102 +1,179 @@
...
   componentWillReceiveProps(nextProps) {
     this.updateChecklistState(nextProps);
   }
 
+  getCompletionCountID = () => (`${this.props.idPrefix.split(/\s/).join('-')}-completion-count`);
+
   getHeading = () => (
     <h3 className={classNames('font-weight-normal', styles['font-extra-large'])}>
-      <span id={this.state.headingID}>{this.props.dataHeading}</span>
+      <span aria-describedby={this.getCompletionCountID()}>{this.props.dataHeading}</span>
```
MichaelRoytman: The `aria-describedby` attribute should be on the header, not the element describing the header, so I have swapped that.

src/utils/CourseChecklist/courseChecklistData.jsx
```diff
@@ -5,72 +5,50 @@ export const filters = {
 };
 
 export const launchChecklist = {
```
MichaelRoytman: I'm not sure how useful this file is anymore. I could just move all this data into the component itself. Thoughts?

src/components/CourseChecklist/CourseChecklist.scss
fysheets: They seem generic enough that moving them up to the top level makes sense to me. You can either move it now, or wait until it's necessary.

src/components/CourseChecklist/CourseChecklist.test.jsx
fysheets: That's a fun function to generate the object.

src/components/CourseChecklist/index.jsx
fysheets: Not vital, but could the aria-describedby go onto the `h3` and then the `span` could be removed?

src/components/CourseChecklist/index.jsx
```diff
@@ -1,102 +1,172 @@
...
+  }
+
+  getCompletionIcon = (checkID) => {
+    const isCompleted = this.isCheckCompleted(checkID);
+    const message = isCompleted ? messages.completedItemLabel : messages.uncompletedItemLabel;
+
+    return (
+      <WrappedMessage message={message}>
+        {displayText =>
+          (<Icon className={[classNames('m-4 col-1', FontAwesomeStyles.fa, FontAwesomeStyles['fa-2x'], this.getCompletionIconClassNames(isCompleted))]} id={`icon-${checkID}`} screenReaderText={displayText} />)
```
fysheets: Nit: Can this be broken onto more than one line so that it's easier to read?

src/components/CourseChecklist/index.jsx
```diff
@@ -1,102 +1,172 @@
...
+  )
+
+  getUpdateLinkDestination = (checkID) => {
+    switch (checkID) {
+      case 'welcomeMessage': return this.props.links.course_updates;
+      case 'gradingPolicy': return this.props.links.grading_policy;
+      case 'certificate': return this.props.links.certificates;
+      case 'courseDates': return `${this.props.links.settings}#schedule`;
+      default: return null;
+    }
```
fysheets: 👍

src/components/CourseChecklist/index.jsx
```diff
@@ -1,102 +1,172 @@
...
+        destination={this.getUpdateLinkDestination(checkID)}
+      />
+    </div>
+  );
+
+  getListItems = () => (
+    this.state.checks.map((check) => {
+      const isCompleted = this.isCheckCompleted(check.id);
+
+      return (<div id={`checklist-item-${check.id}`} className={classNames('bg-white border my-1 py-4', this.getChecklistItemColorClassName(isCompleted), styles['checklist-item'])} key={check.id}>
```
fysheets: Same nit about line length on this one

src/utils/CourseChecklist/courseChecklistData.jsx
fysheets: Which component were you thinking of moving it into? Do you think it would be easier to understand if it were to be moved there?

PR #233, opened 2018-07-12
mroytman/checklist loading state

MichaelRoytman:


src/components/CourseChecklist/displayMessages.jsx
```diff
@@ -131,6 +131,11 @@ const messages = defineMessages({
     defaultMessage: 'uncompleted',
     description: 'Label that describes an uncompleted task',
   },
+  loadingChecklistLabel: {
```
MichaelRoytman: I was thinking perhaps we'd want "loading course launch checklist" instead or something, but I think that because the checklist has the header above the loading icon, it would be okay to leave it this way.

src/components/CourseOutlineStatus/displayMessages.jsx
```diff
@@ -0,0 +1,11 @@
+import { defineMessages } from 'react-intl';
+
+const messages = defineMessages({
+  loadingChecklistLabel: {
+    id: 'loadingChecklistLabel',
+    defaultMessage: 'Loading',
```
MichaelRoytman: In this case, I wonder if this should be a little more explicit. Maybe "loading checklist score" or something?

src/components/CourseChecklist/displayMessages.jsx
fysheets: Hm I see what you're saying. I wonder if it would it make sense to have the Icon be hidden and have the header read out: "headerText loading" if it's loading / "loaded" once the data has been rendered? Or do you think that would be too wordy? 

Same goes for the course outline & the Checklists heading there.

PR #239, opened 2018-07-24
use ora validation

sanfordstudent:


src/utils/CourseChecklist/courseChecklistValidators.js
```diff
@@ -23,6 +23,10 @@ export const hasAssignmentDeadlines = (assignments, dates) => {
     return false;
   } else if (assignments.assignments_with_dates_after_end.length > 0) {
     return false;
+  } else if (assignments.assignments_with_ora_dates_after_end.length > 0) {
```
MichaelRoytman: If you could add two test cases for each of these conditional statements in `src/utils/CourseChecklist/courseChecklistValidators.test.js`, that would be great.

PR #247, opened 2018-08-29
Do not show certificate check when certificates url is empty

awaisdar001:


src/components/CourseChecklist/index.jsx
```diff
@@ -173,6 +173,11 @@ class CourseChecklist extends React.Component {
     this.state.checks.map((check) => {
       const isCompleted = this.isCheckCompleted(check.id);
       const shouldShowCommentSection = this.shouldShowCommentSection(check.id);
+      const shouldThisShowCheck = this.shouldShowCheck(check.id);
```
MichaelRoytman: Nit:  could you rename this variable to shouldShowCheck?

src/components/CourseChecklist/index.jsx
```diff
@@ -327,6 +332,13 @@ class CourseChecklist extends React.Component {
     return false;
   }
 
+  shouldShowCheck = (checkID) => {
+    if (checkID === 'certificates') {
+      return this.props.studioDetails.links.certificates.trim() !== '';
+    }
```
MichaelRoytman: Could you also add a condition for weekly highlights, since we do not want weekly highlights to show in the course checklist either. You will probably have to make changes to the course_quality API in edx-platform in order to pass through some kind of flag in the JSON response that indicates whether weekly highlights are enabled.

src/components/CourseChecklist/index.jsx
```diff
@@ -327,6 +332,13 @@ class CourseChecklist extends React.Component {
     return false;
   }
 
+  shouldShowCheck = (checkID) => {
+    if (checkID === 'certificates') {
+      return this.props.studioDetails.links.certificates.trim() !== '';
```
MichaelRoytman: This condition will have to change based on the feedback Farhanah gave you on https://github.com/edx/edx-platform/pull/18842 in order to check for something like `this.props.data.certificates.is_enabled` instead.

src/components/CourseChecklist/index.jsx
```diff
@@ -173,6 +173,11 @@ class CourseChecklist extends React.Component {
     this.state.checks.map((check) => {
       const isCompleted = this.isCheckCompleted(check.id);
       const shouldShowCommentSection = this.shouldShowCommentSection(check.id);
+      const shouldThisShowCheck = this.shouldShowCheck(check.id);
+
+      if (!shouldThisShowCheck) {
```
MichaelRoytman: You could also make the condition positive by checking

```
if (shouldShowCheck) {
  return (...);
} else {
  return null;
}
```

That might be easier to read.

src/components/CourseChecklist/CourseChecklist.test.jsx
```diff
@@ -164,6 +164,20 @@ describe('CourseChecklist', () => {
         expect(listItems).toHaveLength(testChecklistData.length);
       });
 
+      it('the correct number of checks if certificates url is empty', () => {
```
MichaelRoytman: The test looks great, but you'll need to change `newProps` once you've made changes to the edx-platform course_validation API that Farhanah recommended.

MichaelRoytman: Also, please add a test for weekly highlights, as well, once you add that to the course checklist logic.

src/components/CourseChecklist/index.jsx
awaisdar001: @MichaelRoytman Next I will focus on weekly highlights after certificate issue is clear.

awaisdar001:
Related PR in Platform: https://github.com/edx/edx-platform/pull/18842

awaisdar001:
@fysheets / @MichaelRoytman : Finally, the build is green, after fixing/adding tests. Can you please take a look now?

src/components/CourseOutlineStatus/index.jsx
```diff
@@ -45,17 +47,22 @@ export default class CourseOutlineStatus extends React.Component {
...
+    let courseData;
     if (Object.keys(nextProps.courseLaunchData).length > 0) {
-      const checks = getFilteredChecklist(
-        launchChecklist.data, nextProps.courseLaunchData.is_self_paced);
+      courseData = {
+        is_self_paced: nextProps.courseLaunchData.is_self_paced,
+        certificates: nextProps.courseLaunchData.certificates,
+      };
+      const filteredCourseLaunchChecks = getFilteredChecklist(launchChecklist.data,
+        { data: courseData });
```
fysheets: Actually - I think I understand not just sending over `this.props` as the lone variable because of calls like this where we're using data that isn't coming from `nextProps` for the first parameter. I'm still weary of calling the second parameter `props` if we're not sending over the React props value as a whole, but I think that I do understand having it be two different parameters.

src/components/CourseChecklist/index.jsx
```diff
@@ -330,7 +330,7 @@ class CourseChecklist extends React.Component {
   updateChecklistState(props) {
     if (Object.keys(props.data).length > 0) {
       const checks = getFilteredChecklist(
-        props.dataList, props.data.is_self_paced);
+        props.dataList, props);
```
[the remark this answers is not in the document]
awaisdar001: @fysheets You are right, `getFilteredChecklist` is used is `CourseCheklist` as well as `CourseOutlineStatus`, with different dataLists.

src/components/CourseOutlineStatus/index.jsx
```diff
@@ -31,11 +31,13 @@ export default class CourseOutlineStatus extends React.Component {
...
   componentWillMount() {
+    const isSelfPaceCourse = this.props.studioDetails.course.is_course_self_paced;
+    const props = { data: { is_self_paced: isSelfPaceCourse } };
     this.setState({
       totalCourseBestPracticesChecks: getFilteredChecklist(bestPracticesChecklist.data,
-        this.props.studioDetails.course.is_course_self_paced).length,
+        props).length,
       totalCourseLaunchChecks: getFilteredChecklist(launchChecklist.data,
-        this.props.studioDetails.course.is_course_self_paced).length,
+        props).length,
```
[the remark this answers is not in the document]
awaisdar001: The reason I didn't pass the whole `this.props` is that data is available in different structure in `CourseOutlineStatus`. e.g. `is_self_paced` available in `this.props.studioDetails.course.is_course_self_paced` while on the otherhand in `CourseChecklist` it is available in `props.data.is_self_paced`.  

Due to this, I had to create the custom data object here to pass.

src/components/CourseOutlineStatus/index.jsx
awaisdar001: I think we can rename the variable `props`.

src/utils/CourseChecklist/courseChecklistValidators.js
```diff
@@ -64,7 +64,7 @@ export const hasDiverseSequences = (subsections) => {
 };
 
 export const hasWeeklyHighlights = sections => (
-  sections.highlights_enabled
+  sections.highlights_active_for_course
```
fysheets: For the logic here, we will want to do `sections.highlights_active_for_course && sections.highlights_enabled` as this validation is checking to see if the course team has turned on weekly highlights in their course. So the check will first validate that highlights are active, then if the course team is using them.

src/components/CourseChecklist/index.jsx
```diff
@@ -329,8 +329,11 @@ class CourseChecklist extends React.Component {
 
   updateChecklistState(props) {
     if (Object.keys(props.data).length > 0) {
-      const checks = getFilteredChecklist(
-        props.dataList, props.data.is_self_paced);
+      const isSelfPaced = props.data.is_self_paced;
+      const hasCertificatesEnabled = props.data.certificates && props.data.certificates.is_enabled;
+      const hasHighlightsEnabled = props.data.sections && props.data.sections.highlights_enabled;
+      const checks = getFilteredChecklist(props.dataList,
+        isSelfPaced, hasCertificatesEnabled, hasHighlightsEnabled);
```
fysheets: This is much easier to follow (same as the OutlineStatus change) - thank you for the change!

awaisdar001:
fixed the tests and made the change requested previously. 
@MichaelRoytman / @fysheets Please review.

awaisdar001:
https://github.com/edx/edx-platform/pull/18842 is merged, and now waiting for it to land on prod before merging this pr.

fysheets:
@awaisdar001 since https://github.com/edx/edx-platform/pull/18842 looks okay on stage and has been in limbo for a couple of days, I think it would be okay to rebase this with the latest data and merge it in. You can then get the follow on PR created in edx-platform and ready to go.

awaisdar001:
Just rebased before I can merge it.

PR #249, opened 2018-09-12
[wip]: Update Search and Pagination on Assets Page & Edit Image Modal to 

matthugs:
This updates the search and pagination components on the assets page (which are shared with the modal used to choose an image for an HTML unit) to use the paragon components provided for this purpose. Consistency!

Still TODO:
- some literal TODO comments in the code
- styling is slightly broken on the assets page (but not the edit image modal /shrug)
- labels missing from pagination b/c of https://github.com/matthugs/trouble-in-intl-land

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^1.0.0",
-    "@edx/paragon": "3.1.2",
+    "@edx/paragon": "3.4.7",
     "airbnb-prop-types": "^2.10.0",
     "babel-polyfill": "^6.26.0",
     "classnames": "^2.2.5",
--- a/src/components/AssetsSearch/AssetsSearch.test.jsx
+++ b/src/components/AssetsSearch/AssetsSearch.test.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { SearchField } from '@edx/paragon';
 
 import AssetsSearch from './index';
 import { searchInitial } from './../../data/reducers/assets';
@@ -10,36 +11,24 @@ const defaultProps = {
   courseDetails: {},
 };
 
-let wrapper;
 
 describe('<AssetsSearch />', () => {
   describe('renders', () => {
+    let wrapper,
+        searchField;
     beforeEach(() => {
       wrapper = mountWithIntl(
         <AssetsSearch
           {...defaultProps}
         />,
       );
+      searchField = wrapper.find(SearchField);
     });
-    it('has label, input, and button in a form element', () => {
-      const searchForm = wrapper.find('form');
-
-      expect(searchForm).toHaveLength(1);
-      expect(searchForm.find('label')).toHaveLength(1);
-      expect(searchForm.find('input[type="search"]')).toHaveLength(1);
-      expect(searchForm.find('button[type="submit"]')).toHaveLength(1);
-    });
-    it('has correct styling', () => {
-      const button = wrapper.find('form').find('button[type="submit"]');
-      expect(wrapper.find('form').hasClass('form-group')).toEqual(true);
-      expect(wrapper.find('form').hasClass('form-inline')).toEqual(true);
-      expect(button.find('span.fa')).toHaveLength(1);
-      expect(button.find('span.fa').hasClass('fa-search')).toEqual(true);
+    it('has a paragon SearchField', () => {
+      expect(searchField).toHaveLength(1);
     });
     it('handles onChange callback correctly', () => {
-      const searchInput = wrapper.find('form input[type="search"]');
-
-      searchInput.simulate('change', { target: { value: 'edX' } });
+      searchField.prop('onChange')('edX');
       expect(wrapper.state('value')).toEqual('edX');
     });
     it('calls updateSearch when submit button is clicked', () => {
@@ -49,19 +38,17 @@ describe('<AssetsSearch />', () => {
         updateSearch: searchSpy,
       });
 
-      const submitButton = wrapper.find('form button[type="submit"]');
-      const searchInput = wrapper.find('form input[type="search"]');
-      searchInput.simulate('change', { target: { value: 'edX' } });
-      submitButton.simulate('submit');
+      searchField.prop('onChange')('edX');
+      searchField.prop('onSubmit')();
 
       expect(searchSpy).toHaveBeenCalledTimes(1);
       expect(searchSpy).toHaveBeenCalledWith('edX', defaultProps.courseDetails);
     });
     it('updates search input text when search redux state is updated', () => {
-      const searchInput = wrapper.find('form input[type="search"]');
       wrapper.setProps({ assetsSearch: { search: 'foobar' } }, () => {
+        searchField = wrapper.find(SearchField);
         expect(wrapper.state('value')).toEqual('foobar');
-        expect(searchInput.instance().value).toEqual('foobar');
+        expect(searchField.prop('value')).toEqual('foobar');
       });
     });
   });
--- a/src/components/AssetsSearch/index.jsx
+++ b/src/components/AssetsSearch/index.jsx
@@ -1,8 +1,6 @@
 import React from 'react';
-import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import { Button, Icon, InputText } from '@edx/paragon';
-import FontAwesomeStyles from 'font-awesome/css/font-awesome.min.css';
+import { SearchField } from '@edx/paragon';
 
 import WrappedMessage from '../../utils/i18n/formattedMessageWrapper';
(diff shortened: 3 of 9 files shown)


src/components/AssetsSearch/AssetsSearch.test.jsx
```diff
@@ -10,36 +11,24 @@ const defaultProps = {
   courseDetails: {},
 };
 
-let wrapper;
 
 describe('<AssetsSearch />', () => {
   describe('renders', () => {
+    let wrapper;
```
fysheets: Out of curiosity, what was the reasoning around moving this down? Just want to make sure that I understand.

src/components/AssetsSearch/index.jsx
```diff
@@ -23,49 +20,28 @@ export default class AssetsSearch extends React.Component {
     }
   }
 
-  submit(event) {
+  submit = () => {
     this.props.updateSearch(this.state.value, this.props.courseDetails);
-    event.preventDefault(); // prevent the page from navigating to the form action URL
```
fysheets: Does the Paragon component handle this behavior for us now?

src/components/AssetsSearch/index.jsx
```diff
@@ -23,49 +20,28 @@ export default class AssetsSearch extends React.Component {
...
-          )}
+          inputLabel={<WrappedMessage message={messages.assetsSearchInputLabel} />}
+          screenReaderText={{
+            clearButton: <WrappedMessage message={messages.assetsClearSearchButtonLabel} />,
+            searchButton: <WrappedMessage message={messages.assetsSearchSubmitLabel} />,
+          }}
+          value={this.state.value}
         />
-      </form>
+      </div>
```
fysheets: Well, that's much cleaner

src/components/Pagination/Pagination.scss
```diff
@@ -29,3 +28,6 @@
     @extend .disabled;
   }
 }
+.page-item.disabled .page-link {
+  background-color: transparent;
+}
```
fysheets: Can you share the context on why this CSS change was needed?

src/components/Pagination/index.jsx
```diff
@@ -1,90 +1,46 @@
...
-            </nav>
+            <ParagonPagination
+              paginationLabel={paginationLabel}
+              pageCount={totalPages}
+              buttonLabels={{
+                previous: this.props.intl.formatMessage(messages.paginationPrevious),
+                next: this.props.intl.formatMessage(messages.paginationNext),
+                page: this.props.intl.formatMessage(messages.paginationPage),
+                curerntPage: this.props.intl.formatMessage(messages.paginationCurrentPage),
+                pageOfCount: this.props.intl.formatMessage(messages.paginationOf),
```
fysheets: Glad we were able to make this work!

src/data/actions/assets.js
```diff
@@ -188,11 +184,6 @@ export const pageUpdate = (page, courseDetails) =>
 
     return dispatch(getAssets(parameters, courseDetails)).then((responseAction) => {
       if (requestFailed(responseAction)) {
-        // react-paginate does not force the page state via the forcePage prop when
-        // the previous and next props for forcePage are the same; therefore, we
-        // have to reset the page state after a failed click in order to be able
-        // to reset the page
-        dispatch(resetPageState());
```
fysheets: Does the Paragon component handle this failure more gracefully?

src/components/AssetsSearch/AssetsSearch.test.jsx
matthugs: The muscle memory: I tend to put variable declarations into the closest possible enclosing scope whenever writing JS.

Effects:
- helps clarify what scope something is needed for
- when adding references to the variable to wider scopes, forces you to acknowledge that that could be a confusing change by forcing you to move the declaration
- (no longer a real problem with ES6 modules) helps avoid accidentally globally-scoped variables

src/components/AssetsSearch/index.jsx
matthugs: Correct. Paragon doesn't expose the event via its onSubmit callback prop, either, so there's nothing available on which to call `preventDefault()`.

src/components/AssetsSearch/index.jsx
matthugs: 🎉

src/components/Pagination/Pagination.scss
matthugs: Yes; it's to override some CSS for making all the list items look similar. The disabled list items in this context are the ellipses representing ranges of pages not visible. The design of the paragon component is that those elements shouldn't have a distinct background color, and should blend in with the surrounding background color.

With that said I am _quite_ bad at CSS/SCSS; is this the use-case for a SCSS variable override? It def looks like I could've put it above in the `.disabled { ... .page-link { <here> } }` above and achieved the same degree of specificity. I'll try that out for my next push.

src/data/actions/assets.js
matthugs: Correct! I tested whether this use-case were necessary using network throttling features: I throttled to slow to make the request slow enough to catch, and then turned off the network while it was in-flight to trigger this code path. After deletion, there was no difference in behavior without it.

src/components/Pagination/index.jsx
fysheets: Cool - Interested to hear about it!

fysheets:
Resolved conversations - thanks for the clarity. Two quick follow ups - please rebase on top of the latest master changes and a quick clarification on what the `refactor` commit message will do for version numbers.

matthugs:
> Resolved conversations - thanks for the clarity. Two quick follow ups - please rebase on top of the latest master changes and a quick clarification on what the `refactor` commit message will do for version numbers.

Good catch, @fysheets!

package.json
```diff
@@ -15,7 +15,7 @@
   "license": "AGPL-3.0",
   "dependencies": {
     "@edx/edx-bootstrap": "^1.0.0",
-    "@edx/paragon": "3.1.2",
+    "@edx/paragon": "3.4.7",
```
matthugs: I do have one final detail to follow-up on this PR which relates to some propType errors I expected, but didn't see. They were stemming from changes in edx/paragon I thought were necessary, but which we also ran into trouble trying to test locally. Following up on that in just a little bit! (It seems like the proptype validations went away when I bumped the paragon version, but my understanding was that I should still expect them even if we were up-to-date with `@edx/paragon`'s `master`...

matthugs: It looks like proptype validations aren't running against paragon components in the new version we're upgrading into. FYI @fysheets, since this may be a larger issue. We may have a number of proptype issues masked by this issue.

I haven't dug in super far, but I expect this may also be the result of how we use post-build paragon code from within studio-frontend.

matthugs: We're moving forward from this with a bumped paragon version, which should suffice to eliminate any proptype validation errors.

fysheets:
Sorry for the back and forth here - Thanks for updating that logic it looks awesome! I was running through it and Michael noticed that the CSS didn't quite match up to what may have been expected. We can sync up on that in person to determine the best course of action there. (Timebox to try to fix in this PR or a follow on task)

matthugs:
> Sorry for the back and forth here - Thanks for updating that logic it looks awesome! I was running through it and Michael noticed that the CSS didn't quite match up to what may have been expected. We can sync up on that in person to determine the best course of action there. (Timebox to try to fix in this PR or a follow on task)

No worries re:back-and-forth @fysheets and @MichaelRoytman! Happy to have close eyes on my first substantive changes

src/components/Pagination/Pagination.scss
```diff
@@ -1,4 +1,7 @@
+$pagination-disabled-bg: transparent;
```
fysheets: Is this another addition due to the var-reset?

fysheets: Actually never mind, I just had to scroll down another 2 inches on the screen

matthugs: No problem; added another commented override in this file to fix the style issue you pointed out in meatspace. I can just TASTE the merge

src/components/Pagination/index.jsx
```diff
@@ -1,90 +1,46 @@
...
-              />
-            </nav>
+            <ParagonPagination
+              paginationLabel={paginationLabel}
+              pageCount={totalPages}
+              buttonLabels={{
+                previous: this.props.intl.formatMessage(messages.paginationPrevious),
+                next: this.props.intl.formatMessage(messages.paginationNext),
+                page: this.props.intl.formatMessage(messages.paginationPage),
+                curerntPage: this.props.intl.formatMessage(messages.paginationCurrentPage),
```
dsjen: Spelling? `curerntPage`

PR #283, opened 2019-12-31
fix(a11ypolicypage): update studio a11y page

awaisdar001:
**Update studio a11y page with the new proposed content.** 

Screenshot after update.
Screenshot 2019-12-31 at 16 23 51
Frontend PR: edx/studio-frontend#283
Platform PR: _pending..._

FYI -- @edx/sustaining-mavericks


fysheets:
I will need to pull this down to test this locally -- I see that you changed some of the variable names and workflows instead of just the text fields. I would like to test it to validate the code review.

src/components/AccessibilityBody/index.jsx
```diff
@@ -35,12 +36,12 @@ const AccessibilityBody = (props) => {
...
           message={messages.a11yBodyEmailHeading}
           values={{
-            emailElement,
+            emailElement: ReactDomServer.renderToString(emailElement),
           }}
         >
           {
             displayText => (
-              <li>{displayText}
+              <li><span dangerouslySetInnerHTML={{ __html: displayText }} />
```
awaisdar001: This is a workaround described in : https://stackoverflow.com/questions/59542379/react-intl-wrapping-child-elements-in-formatmessage?noredirect=1#comment105255145_59542379


If you are certain about the code you want to insert is **not** malicious it's a pretty safe bet to use `dangerouslySetInnerHTML`. If you trust the API you have little to worry about. For example, this is perfectly safe:
```
return <div dangerouslySetInnerHTML={{__html: "<p>foo bar</p>"}} />;
```

awaisdar001:
> I will need to pull this down to test this locally -- I see that you changed some of the variable names and workflows instead of just the text fields. I would like to test it to validate the code review.

thank you!

awaisdar001:
@fysheets or @MichaelRoytman if you get a chance, please do review this PR.

src/components/AccessibilityBody/index.jsx
fysheets: @awaisdar001 Why is this necessary? Why could we not leave this the way it was before?

src/components/AccessibilityBody/index.jsx
```diff
@@ -71,52 +72,7 @@ const AccessibilityBody = (props) => {
           message={messages.a11yBodyExtraInfo}
           tagName="li"
         />
-        <WrappedMessage
```
fysheets: Why were these blocks removed? Were they removed in the updated text we were given?

awaisdar001: Correct. They were removed in the updated text that is given. 

With the new template of studio-a11y, these text blocks were removed and no longer needed to display. 
See 
https://studio.edx.org/accessibility
vs
Test

src/components/AccessibilityBody/index.jsx
awaisdar001: @fysheets Currently, we are displaying information correctly on the studio-a11y page.
We are supposed to display 


> Notify us by email addressed to [accessibility@edx.org](mailto:accessibility@edx.org).  In order for edX to fully and promptly assess and respond to your request, please provide the following information:


but due to this bug, we instead display (See: https://studio.edx.org/accessibility)

> Notify us by email addressed to


This is why I have to take this workaround which is an issue in `react-intl`. 


if we leave this the way it was before it will be rendered as 

> Send an email to

where it should 

> Send an email to [accessibility@edx.org](ddd) with the following information:

I have spent the time to make it work which looks elegant in the backend, but in the end, the only workaround I had was this. 
Here is the sandbox: 
https://codesandbox.io/s/formattedmessage-render-props-c6hfn

awaisdar001:
@fysheets Ping :)

src/components/AccessibilityBody/index.jsx
fysheets: @awaisdar001 As discussed in person, thank you for the clarity can you please add a comment to this file to briefly explain this decision so that future editors don't have to find this PR for context.

src/components/AccessibilityBody/index.jsx
```diff
@@ -35,12 +36,15 @@ const AccessibilityBody = (props) => {
...
-            emailElement,
+            emailElement: ReactDomServer.renderToString(emailElement),
           }}
         >
           {
             displayText => (
-              <li>{displayText}
+              // I am using dangerouslySetInnerHTML to as workaround to fix an issue that does not render
+              // html element in the parent when it has its own children.
+              // I am certain about the string being inserted here is not malicious so it's a pretty safe bet to use dangerouslySetInnerHTML.
```
awaisdar001: @fysheets Sorry for the delay just added the comment.

PR #315, opened 2022-01-17
feat: Added support for Node 16

Jawayria:
Added support for node 14 and node 16. I had to make some additional changes to ensure successful tests execution and tobuild docker image successfully on node 16.

**Additional Changes**

- Added options in Webpack config to fix	`Syntax Error: Support for the experimental syntax 'jsx' isn't currently enabled in react js`
- Updated babel-loader version to fix `module build failed (from ./node_modules/babel-loader/lib/index.js) syntaxerror unexpected token`
- Updated Webpack-cli version to fix `describe: optionsSchema.definitions.output.properties.path.description with webpack 4.8.3`
- Replaced node-sass with sass to fix `Error: 'node-sass' version 6.0.0 is incompatible with ^4.0.0`
- Updated redux and redux-devtools-extension version to fix conflict with latest redux-thunk version


.github/workflows/ci.yml
```diff
@@ -15,10 +15,13 @@ jobs:
     - name: Checkout
       uses: actions/checkout@v2
 
-    - name: Setup Nodejs
-      uses: actions/setup-node@v1
+    - name: Setup Node.js Env
+      run: echo "NODE_VER=`cat .nvmrc`" >> $GITHUB_ENV
```
mraarif, line 19: why don't we specify node-js versions in under `matrix`

.github/workflows/ci.yml
```diff
@@ -42,15 +45,15 @@ jobs:
...
-      with:
-        server_address: email-smtp.us-east-1.amazonaws.com
-        server_port: 465
-        username: ${{secrets.EDX_SMTP_USERNAME}}
-        password: ${{secrets.EDX_SMTP_PASSWORD}}
-        subject: CI workflow failed in ${{github.repository}}
-        to: teaching-and-learning@edx.opsgenie.net
-        from: github-actions <github-actions@edx.org>
-        body: CI workflow in ${{github.repository}} failed! For details see "github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+#    - name: Send failure notification
```
mraarif, line 48: why is the failure notification disabled?

.github/workflows/npm-publish.yml
```diff
@@ -14,10 +14,13 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Setup Node.js Env
+        run: echo "NODE_VER=`cat .nvmrc`" >> $GITHUB_ENV
+
       - name: Setup Node.js
-        uses: actions/setup-node@v1
+        uses: dcodeIO/setup-node-nvm@v4
```
mraarif, line 21: is there a problem with `actions/setup-node` setting up 16

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:8.11
+FROM node:16.13
```
mraarif, line 1: did you build the docker image to verify?

package.json
```diff
@@ -65,7 +65,7 @@
     "identity-obj-proxy": "^3.0.0",
     "ignore-loader": "^0.1.2",
     "jest": "^22.4.3",
-    "node-sass": "^4.12.0",
+    "node-sass": "^6.0.1",
```
mraarif, line 68: does this version upgrade works with both Node versions 12 and 16

Jawayria: Yes it works with both

.github/workflows/ci.yml
Jawayria: That's a temporary change., just to avoid unnecessary notifications until the PR gets finalised

.github/workflows/ci.yml
```diff
@@ -11,14 +11,18 @@ on:
 jobs:
   build:
```
mraarif, line 12: shouldn't this be named `tests`? so the workflow jobs are named like `CI/Node:12-tests`

.github/workflows/npm-publish.yml
```diff
@@ -17,7 +21,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: ${{ matrix.node }}
```
aht007, line 24: In my opinion we should only be publishing the package with the stable node version.

.github/workflows/ci.yml
```diff
@@ -9,19 +9,23 @@ on:
...
 
     - name: Setup Nodejs
       uses: actions/setup-node@v1
       with:
-        node-version: 12
+        node-version: ${{ matrix.node }}
 
     - name: Setup npm
-      run: npm i -g npm@5.8.0
+      run: npm i -g npm@8.1.2
```
mraarif, line 28: can we just specify the major version and let `npm` install any latest minor versions

.github/workflows/ci.yml
mraarif: we should enable this now that there are no failures

.github/workflows/npm-publish.yml
mraarif: yes, we should have only one version of Node publishing this package, i.e 12, later we can drop older versions and set it to 16

Dockerfile
mraarif: we should only specify the major version of nodejs here and let docker pull any latest tags

Dockerfile
```diff
@@ -1,13 +1,20 @@
...
 
 USER root
 RUN apt-get update
 RUN apt-get install -y vim
-RUN npm i -g npm@6.1.0
+RUN npm i -g npm@8.1.2
```
mraarif, line 6: same as above, we should only specify the major version

aht007 (approved):

PR #336, opened 2022-04-05
build: update push translation job

DawoudSheraz:
### Description

- Updating Studio-frontend push translation job to match the other MFEs push translation job
- This is needed to update push translation jobs to use edx/reactifex instead of reactifex. edx/reactifex uses Transifex API v3 for pushing translation string comments instead of API v2.
- In push job, `tx push -s` has been added. Not sure how the translation strings are pushed in studio frontend. The comments suggest that transifex_input.json is being by Transifex but I could not find any reference on [Transifex](https://www.transifex.com/open-edx/edx-platform/studio-frontend/)
- References
   - https://github.com/openedx/frontend-platform/blob/master/src/i18n/scripts/Makefile
   - https://github.com/openedx/frontend-platform/blob/b5f6b0847a2ff774b8bcb9664b2789b0fa9511a8/docs/how_tos/i18n.rst
   - https://github.com/openedx/frontend-app-support-tools/blob/master/Makefile
   - https://github.com/openedx/frontend-build/blob/a8b528da3b8398bfb27935386932a7fca98c7a19/config/babel.config.js

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -15,7 +15,7 @@
     "i18n": {
       "plugins": [
         ["react-intl", {
-          "messagesDir": "./src/data/i18n/default/",
+          "messagesDir": "./temp/babel-plugin-react-intl",
           "enforceDescriptions": true
         }]
       ]
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,13 @@
 node_modules
 npm-debug.log
 coverage
+.idea
 
 dist/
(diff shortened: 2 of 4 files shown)


ansabgillani (approved):

davidjoy (approved):
This looks reasonable to me.

connorhaugh (approved):

PR #372, opened 2023-03-03
Remove unused tranisfex v2 url 

Mashal-m:
Ticket:
[Updating /api/v2 transifex endpoints](https://github.com/edx/edx-arch-experiments/issues/202)

There was no need for tx-url1 and tx-url2 because transifex has already migrated from v2 to v3.


UsamaSadiq (approved):

schenedx (approved):
👍

PR #379, opened 2023-05-09
refactor: Don't specify the settings file.

feanil:
The edx-platform `manage.py` file defaults to the devstack_docker file
by default. We don't need to specify it here.

Note: The motivation for removing this is that it's weird coupling to edx-platform that makes it harder to change the names or re-organize settings files over there.  We'd like to be able to re-organize/rename settings files there without breaking other downstream consumers.


kdmccormick (approved):
> Note: The motivation for removing this is that it's weird coupling to
edx-platform that makes it harder to change the names or re-organize
settings files over there.  We'd like to be able to re-organize/rename
settings files there without breaking other downstream consumers.

👍🏻👍🏻👍🏻

PR #404, opened 2025-09-26
build: Remove reactifex dependency.

feanil:
This repo only relies on that deprecated library for one script and
since this repo is also on the road to deprecation, for now I've just
copied the one script we needed from that repo to here so that we can
drop that dependency and archive that repo.


Makefile
```diff
@@ -2,7 +2,7 @@ UNAME := $(shell uname)
 
 i18n = ./src/i18n
 transifex_input = $(i18n)/transifex_input.json
-transifex_utils = ./node_modules/.bin/edx_reactifex
+transifex_utils = node ./scripts/concat_translations.js
```
arbrandes, line 5: This doesn't look like [a regular transifex_utils definition](https://github.com/openedx/frontend-app-account/blob/master/Makefile#L2).  Any particular reason why?

brian-smith-tcril (approved):
LGTM!

I'm on board with the "just copy the script we need in for now" solution, and I assume you've verified that the `transifex_input.json` file this generates with these changes matches the pre-change version.

feanil:
Yep, the output is the same from my testing.

